PR #1, opened 2018-08-08
Modify the regex statement to work on MacGPG

idrysdale:
When running `gpg --version` on a mac, it outputs:

```
gpg (GnuPG/MacGPG2) 2.2.8
libgcrypt 1.8.3
Copyright (C) 2018 Free Software Foundation, Inc.
```

This didn't fit the regex we were using.

### DIFF
--- a/src/fluidkeys.go
+++ b/src/fluidkeys.go
@@ -15,7 +15,7 @@ func main() {
 	}
 	outString := string(out)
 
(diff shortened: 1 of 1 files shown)


idrysdale:
Check it out. Even had a go at Regex ;-)

paulfurley:
I think this would be a safer regex `gpg \(GnuPG.*\) (\d+\.\d+\.\d+)`

idrysdale:
You're right. Fixed!

PR #18, opened 2018-08-17
Refactoring testing gpgwrapper

idrysdale:


paulfurley:
Hint: the reason this is failing is todo with `.gnupg/gpg.conf` and the `keyid-format` setting differing between your dev machine and the default on travis:

```
?   	github.com/fluidkeys/fluidkeys/colour	[no test files]
ok  	github.com/fluidkeys/fluidkeys/fluidkeys	0.008s
--- FAIL: TestRunGpgWithStdin (0.01s)
    --- FAIL: TestRunGpgWithStdin/with_a_valid_public_key (0.01s)
    	gpgwrapper_test.go:129: GPGs output 'gpg: key 5B85C8D3: public key "test@example.com" imported
    		gpg: Total number processed: 1
    		gpg:               imported: 1  (RSA: 1)
    		' did not contain any success messages
```

Notice that `5B85C8D3` is a short ID, with no preceding `0x`

paulfurley:
(Enhancement: we'll want to control the environment `gpg` runs in more carefully, including writing out our own `gpg.conf`, passing arguments like `--batch` and `--no-tty` etc etc. See https://github.com/paulfurley/expirybot/blob/master/expirybot/gpg_wrapper/sign_text as an example)

idrysdale:
Thanks! Good hint. I've made a start at that by creating `func appendGlobalArguments(arguments ...string)` in `gpgwrapper`.

gpgwrapper/gpgwrapper.go
```diff
@@ -4,12 +4,20 @@ package gpgwrapper
...
 	"errors"
+	"fmt"
+	"io"
 	"os/exec"
 	"regexp"
 )
 
 const GpgPath = "gpg"
 
+var ErrNoVersionStringFound = errors.New("version string not found in GPG output")
```
paulfurley: could that be `const` ?

gpgwrapper/gpgwrapper_test.go
```diff
@@ -1,35 +1,164 @@
...
+
+		if !sawSuccessMessage {
+			t.Errorf("GPGs output '%s' did not contain any success messages", output)
+		}
+	})
 }
 
-func assert_parses_equal(t *testing.T, gpgOutput string, expected string) {
-	actual, err := parseVersionString(gpgOutput)
+func assert_parses_version_correctly(t *testing.T, gpgOutput string, want string) {
```
paulfurley: `assertParsesVersionCorrectly` :)

gpgwrapper/gpgwrapper.go
paulfurley: I tried this and couldn't get it to work

gpgwrapper/gpgwrapper.go
```diff
@@ -4,12 +4,20 @@ package gpgwrapper
...
+
+func ErrProblemExecutingGPG(arguments ...string) error {
```
paulfurley: Cunning

paulfurley: I think it would be incredibly helpful if the error message *also* contained the contents of `stdin` and `stderr` otherwise you can't see what GnuPG complained about when it bombed out

idrysdale:
Ok, excellent, I've made those changes! 👍

PR #20, opened 2018-08-22
Output private key encrypted with password

paulfurley:
* Rewrite PgpKey class
  -  make PgpKey extend openpgp.Entity
  -  remove PgpKey.PublicKey
  -  add Armor and ArmorPrivate methods
  -  pass errors back up from PgpKey.Generate()

* Use github.com/fluidkeys/crypto fork


fluidkeys/main.go
```diff
@@ -31,9 +31,14 @@ func (d DicewarePassword) AsString() string {
 	return strings.Join(d.words, d.separator)
 }
 
+type job struct {
```
idrysdale: Could we better define `job`? Is it `generatePgpKey`?

idrysdale: Ah no - there's a function called that.... 🤔

pgpkey/pgpkey.go
```diff
@@ -17,25 +16,61 @@ const (
...
+	openpgp.Entity
 }
 
-func Generate(email string) PgpKey {
-	config := &packet.Config{RSABits: 4096}
+func Generate(email string) (*PgpKey, error) {
+	return generateKeyOfSize(email, RsaSizeSecureKeyBits)
+}
+
+func GenerateInsecure(email string) (*PgpKey, error) {
```
idrysdale: Can we make this private?

fluidkeys/main.go
paulfurley: renamed to `generatePgpKeyResult`

pgpkey/pgpkey.go
paulfurley: Done

PR #21, opened 2018-08-22
Import key into gpg

idrysdale:


fluidkeys/main.go
```diff
@@ -82,7 +82,8 @@ func main() {
 	}
 	fmt.Println(publicKey)
 	fmt.Println(privateKey)
-	// TODO: use gpgwrapper to import the keys into GnuPG
+	gpgwrapper.ImportArmoredPublicKey(publicKey)
+	fmt.Println("Your public key has been imported into GPG")
```
paulfurley: How about adding a line:
```
To show the key in GnuPG
  > gpg --list-keys '<email>'
```

paulfurley:
:heart:

PR #22, opened 2018-08-22
Create key backup file ~/.fluidkeys/backup.zip

paulfurley:


backupzip/backupzip.go
```diff
@@ -0,0 +1,61 @@
...
+	"os"
+	"path/filepath"
+)
+
+func OutputZipBackupFile(fluidkeysDir, armoredPublicKey string, armoredPrivateKey string) (filename string, err error) {
+	filename = getZipFilename(fluidkeysDir)
+
+	backupZipFile, err := os.Create(filename)
+	if err != nil {
+		return "", err
```
idrysdale: Can we prefix the error with something that offers a little more context in case of failing?

backupzip/backupzip.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	defer backupZipFile.Close()
+
+	err = WriteZipFile(backupZipFile, armoredPublicKey, armoredPrivateKey)
+	if err != nil {
+		return "", err
```
idrysdale: Ditto!

backupzip/backupzip.go
```diff
@@ -0,0 +1,61 @@
+package backupzip
+
+import (
+	"archive/zip"
+	"io"
+	"os"
+	"path/filepath"
+)
+
```
idrysdale: Let's try and get into the habit of commenting our public functions with descriptions of what they do.

PR #24, opened 2018-08-22
Import the private key into GnuPG, too

idrysdale:


paulfurley:
Now we're actually changing GPG's state (importing etc) I think we need to manipulate the home directory so the tests don't mess around with our own `.gnupg` directories :\

PR #32, opened 2018-08-23
Jenkins: clone repo into Go workspace

paulfurley:
1. set GOPATH to a temp directory
2. clone fluidkeys into GOPATH/src/github.com/fluidkeys/fluidkeys

When `script/publish-latest_tag` runs, it makes a fresh `git clone` of
the repo, then `cd`s there and runs `make` etc.

Because of the way Go workspaces work, if we want to import local
packages (e.g. `backupzip`, `pgpkey` etc) by doing:

```
import "github.com/fluidkeys/fluidkeys/pgpkey"
```

... then our own `fluidkeys` repo *must* be checked out into
`$GOPATH/src/github.com/fluidkeys/fluidkeys`

### DIFF
--- a/script/publish_latest_tag
+++ b/script/publish_latest_tag
@@ -1,7 +1,8 @@
 #!/bin/sh -eux
 
 WORKSPACE=${WORKSPACE:-$PWD}
-REPO_DIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


idrysdale:
Very good!

PR #52, opened 2018-09-10
Introduce `key from-gpg`

idrysdale:
We list the secret keys in GPG, then prompt the user to select which to import.

At present this just prints it back.

### DIFF
--- a/fluidkeys/main.go
+++ b/fluidkeys/main.go
@@ -56,6 +56,7 @@ func main() {
 
 Usage:
 	fk key create
+	fk key from-gpg
 
 Options:
 	-h --help    Show this screen`, Version)
@@ -91,9 +92,11 @@ func initSubcommand(args docopt.Opts) exitCode {
 }
 
 func keySubcommand(args docopt.Opts) exitCode {
-	switch getSubcommand(args, []string{"create"}) {
+	switch getSubcommand(args, []string{"create", "from-gpg"}) {
 	case "create":
 		os.Exit(keyCreate())
+	case "from-gpg":
+		os.Exit(keyFromGpg())
 	}
 	panic(fmt.Errorf("keySubcommand got unexpected arguments: %v", args))
 }
--- a/fluidkeys/main.go
+++ b/fluidkeys/main.go
@@ -254,11 +254,11 @@ func promptForKeyToImportFromGpg(secretKeyListings []gpgwrapper.SecretKeyListing
 	for validInput := false; !validInput; {
 		rangePrompt := colour.LightBlue(fmt.Sprintf("[1-%v]", len(secretKeyListings)))
 		input := promptForInput(fmt.Sprintf(PromptWhichKeyFromGPG + " " + rangePrompt + " "))
-		if intergerSelected, err := strconv.Atoi(input); err != nil {
+		if integerSelected, err := strconv.Atoi(input); err != nil {
 			fmt.Print(invalidEntry)
 		} else {
-			if (intergerSelected >= 1) && (intergerSelected <= len(secretKeyListings)) {
-				selectedKey = intergerSelected - 1
+			if (integerSelected >= 1) && (integerSelected <= len(secretKeyListings)) {
+				selectedKey = integerSelected - 1
 				validInput = true
 			} else {
 				fmt.Print(invalidEntry)
--- a/fluidkeys/main.go
+++ b/fluidkeys/main.go
@@ -231,14 +231,22 @@ func generatePgpKey(email string, channel chan generatePgpKeyResult) {
 
 func listKeysForImportingFromGpg(secretKeyListings []gpgwrapper.SecretKeyListing) string {
 	str := fmt.Sprintf("Found %s in GnuPG:\n\n", humanize.Plural(len(secretKeyListings), "key", "keys"))
-	for _, key := range secretKeyListings {
-		str += printSecretKeyListing(key)
+	for index, key := range secretKeyListings {
+		str += printSecretKeyListing(index+1, key)
 	}
 	return str
 }
 
-func printSecretKeyListing(key gpgwrapper.SecretKeyListing) string {
-	output := fmt.Sprintf("    %s\n", key.Fingerprint)
+func printSecretKeyListing(listNumber int, key gpgwrapper.SecretKeyListing) string {
+	runeLengthListNumber := len(strconv.Itoa(listNumber) + ".")
+	remainingIndents := 4 - runeLengthListNumber
+	remainingIndent := ""
+	for i := 0; i < remainingIndents; i++ {
+		remainingIndent += " "
+	}
+	formattedListNumber := colour.LightBlue(strconv.Itoa(listNumber) + ".")
+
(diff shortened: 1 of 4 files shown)


fluidkeys/main.go
```diff
@@ -238,6 +247,25 @@ func printSecretKeyListing(key gpgwrapper.SecretKeyListing) string {
...
+		rangePrompt := colour.LightBlue(fmt.Sprintf("[1-%v]", len(secretKeyListings)))
+		input := promptForInput(fmt.Sprintf(PromptWhichKeyFromGPG + " " + rangePrompt + " "))
+		if intergerSelected, err := strconv.Atoi(input); err != nil {
+			fmt.Printf("Only numbers please\n")
+		} else {
+			if (intergerSelected >= 1) && (intergerSelected <= len(secretKeyListings)) {
+				selectedKey = intergerSelected - 1
+				validInput = true
+			} else {
+				fmt.Printf("Number not available\n")
```
paulfurley: image

"Number not available" felt a bit raw!
* a better way of saying that might be, "please select between 1 and 3"
* it seems to be missing a fullstop
* it might be good to re-display the keys?

fluidkeys/main.go
```diff
@@ -238,6 +247,25 @@ func printSecretKeyListing(key gpgwrapper.SecretKeyListing) string {
 	return output
 }
 
+func promptForKeyToImportFromGpg(secretKeyListings []gpgwrapper.SecretKeyListing) gpgwrapper.SecretKeyListing {
+	var selectedKey int
+	for validInput := false; !validInput; {
+		rangePrompt := colour.LightBlue(fmt.Sprintf("[1-%v]", len(secretKeyListings)))
+		input := promptForInput(fmt.Sprintf(PromptWhichKeyFromGPG + " " + rangePrompt + " "))
+		if intergerSelected, err := strconv.Atoi(input); err != nil {
```
paulfurley: `interger` -> `integer`

fluidkeys/main.go
```diff
@@ -238,6 +247,25 @@ func printSecretKeyListing(key gpgwrapper.SecretKeyListing) string {
...
+			fmt.Printf("Only numbers please\n")
```
paulfurley: Ditto my comment about "Number not available", it's strange feeling error message :)

fluidkeys/main.go
```diff
@@ -231,14 +231,22 @@ func generatePgpKey(email string, channel chan generatePgpKeyResult) {
...
 	return str
 }
 
-func printSecretKeyListing(key gpgwrapper.SecretKeyListing) string {
-	output := fmt.Sprintf("    %s\n", key.Fingerprint)
+func printSecretKeyListing(listNumber int, key gpgwrapper.SecretKeyListing) string {
+	runeLengthListNumber := len(strconv.Itoa(listNumber) + ".")
+	remainingIndents := 4 - runeLengthListNumber
+	remainingIndent := ""
+	for i := 0; i < remainingIndents; i++ {
```
paulfurley: I *think* this is re-implementing https://golang.org/pkg/strings/#Repeat .. isn't it?

fluidkeys/main.go
```diff
@@ -231,14 +231,22 @@ func generatePgpKey(email string, channel chan generatePgpKeyResult) {
...
+		remainingIndent += " "
+	}
+	formattedListNumber := colour.LightBlue(strconv.Itoa(listNumber) + ".")
+
+	output := fmt.Sprintf(formattedListNumber+remainingIndent+"%s\n", key.Fingerprint)
```
paulfurley: I think `fmt.Sprintf("%s%s%s\n, formattedListNumber, remainingIndex, key.Fingerprint)` would be truer to `Sprintf`'s purpose :)

paulfurley:
Awesome to see this come to life, I nearly fell over when I saw my actual keys from `gpg` ! Yay!

fluidkeys/main.go
```diff
@@ -231,14 +231,22 @@ func generatePgpKey(email string, channel chan generatePgpKeyResult) {
...
+	for index, key := range secretKeyListings {
+		str += printSecretKeyListing(index+1, key)
 	}
 	return str
 }
 
-func printSecretKeyListing(key gpgwrapper.SecretKeyListing) string {
-	output := fmt.Sprintf("    %s\n", key.Fingerprint)
+func printSecretKeyListing(listNumber int, key gpgwrapper.SecretKeyListing) string {
+	runeLengthListNumber := len(strconv.Itoa(listNumber) + ".")
```
paulfurley: I think you might be able to replace all this code with a fixed-width format specifier like this:
```
fmt.Printf("%-5sblah\n", "1.")
fmt.Printf("%-5sblah\n", "99.")
```
The `-5` means "5-characters wide, left aligned"
https://play.golang.org/p/CbXqrbuytmQ

fluidkeys/main.go
idrysdale: Yes! Great spot - thanks!

fluidkeys/main.go
idrysdale: Yeah, prefer your suggestion. Will implement that. I think printing the keys again is a bit much given they're just a line or two above.

fluidkeys/main.go
idrysdale: 👍

humanize/plural.go
```diff
@@ -5,13 +5,15 @@ import (
 	"strconv"
 )
 
-func PluralWord(quantity int, singular, plural string) string {
+// Plural returns the quantity then either the `singular` (fox) or `plural`
+// (foxes) depending on that given `quantity`
+// e.g. `Plural(2, "fox", "foxes")` returns "2 foxes"
+func Plural(quantity int, singular, plural string) string {
```
paulfurley: Just a thought, losing the separation of `Plural` and `PluralWord` is a shame, because there's now no way of doing, for example, "no foxes" or "three foxes". I guess what i meant was that it wasn't clear what `Plural` was doing (since it's a noun). *Pluralize* would be better. `PluralWord` made total sense.

paulfurley: That said, happy to crack on and address this when it's needed. From a general style point of view, we should try and use verbs at the start of functions I reckon.

fluidkeys/main.go
```diff
@@ -238,15 +238,8 @@ func formatListedKeysForImportingFromGpg(secretKeyListings []gpgwrapper.SecretKe
...
-	runeLengthListNumber := len(strconv.Itoa(listNumber) + ".")
-	remainingIndents := 4 - runeLengthListNumber
-	remainingIndent := ""
-	for i := 0; i < remainingIndents; i++ {
-		remainingIndent += " "
-	}
-	formattedListNumber := colour.LightBlue(strconv.Itoa(listNumber) + ".")
-
-	output := fmt.Sprintf(formattedListNumber+remainingIndent+"%s\n", key.Fingerprint)
+	formattedListNumber := colour.LightBlue(fmt.Sprintf("%-4s", (strconv.Itoa(listNumber) + ".")))
```
paulfurley: yay

humanize/plural.go
idrysdale: Let's refactor it when we need it. Good idea about verbs.

PR #59, opened 2018-09-11
Prompt [Y/n] when linking key if only one found in GPG

idrysdale:


paulfurley:
Mergefail! :(

fluidkeys/main.go
```diff
@@ -115,17 +115,21 @@ func keyFromGpg() exitCode {
...
-	fmt.Printf("Key to import: %v", keyToImport.Fingerprint)
+	if keyToImport != nil {
+		fmt.Printf("Key to import: %v", keyToImport.Fingerprint)
 
-	fluidkeysDirectory, err := getFluidkeysDirectory()
-	if err != nil {
-		fmt.Printf("Failed to get fluidkeys directory")
-		return 1
-	}
+		fluidkeysDirectory, err := getFluidkeysDirectory()
```
paulfurley: tabs + spaces, make yer mind up guv!

fluidkeys/main.go
```diff
@@ -115,15 +115,21 @@ func keyFromGpg() exitCode {
 	fmt.Printf(formatListedKeysForImportingFromGpg(secretKeys))
 	keyToImport := promptForKeyToImportFromGpg(secretKeys)
 
-	fluidkeysDirectory, err := getFluidkeysDirectory()
-	if err != nil {
-		fmt.Printf("Failed to get fluidkeys directory")
-		return 1
-	}
+	if keyToImport != nil {
```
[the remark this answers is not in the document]
idrysdale: Good shout

idrysdale:
Are you done reviewing this @paulfurley ?

I've made some changes in line with the review.

paulfurley:
Not finished no

fluidkeys/main.go
```diff
@@ -254,24 +259,57 @@ func printSecretKeyListing(listNumber int, key gpgwrapper.SecretKeyListing) stri
...
 			}
 		}
+		return &secretKeyListings[selectedKey]
+	}
+}
+
+func promptToConfirmImportKeyFromGpg(key gpgwrapper.SecretKeyListing) bool {
+	var confirmed bool
+	for validInput := false; !validInput; {
+		input := promptForInput("Import key? [Y/n] ")
```
paulfurley: This advertises a default of `Y` (by virtue of it being uppercase) but it *appears* the switch doesn't support just hitting `[return]`, is that right?

fluidkeys/main.go
```diff
@@ -254,24 +259,57 @@ func printSecretKeyListing(listNumber int, key gpgwrapper.SecretKeyListing) stri
...
+		switch input {
+		case "Y":
```
paulfurley: Any reason why not `switch strings.Lower(input)`?

fluidkeys/main.go
```diff
@@ -254,24 +259,57 @@ func printSecretKeyListing(listNumber int, key gpgwrapper.SecretKeyListing) stri
...
+			validInput = true
+		case "n":
+			confirmed = false
+			validInput = true
+		default:
+			fmt.Printf("Please select only Y or N.\n")
+		}
 	}
-	return secretKeyListings[selectedKey]
+	return confirmed
```
paulfurley: You could lose 2 bits of state and massively simplify this by simply returning:
```
switch stringsToLower(input) {
case "y":
    return true
case "n":
    return false
default:
   ...
}
```

fluidkeys/main.go
idrysdale: Shit, good spot!

fluidkeys/main.go
idrysdale: 👍

paulfurley:
All done!

fluidkeys/main.go
idrysdale: That's nice.

idrysdale:
Cool thanks for that. Some good pointers in there.

I've made changes in line with your suggestions.

PR #62, opened 2018-09-11
`fk key from-gpg`: filter out keys already imported

paulfurley:


paulfurley:
Rebasing

fluidkeys/main.go
```diff
@@ -105,23 +106,39 @@ func keySubcommand(args docopt.Opts) exitCode {
...
-	fmt.Printf(formatListedKeysForImportingFromGpg(secretKeys))
-	keyToImport := promptForKeyToImportFromGpg(secretKeys)
 
-	fluidkeysDirectory, err := getFluidkeysDirectory()
+	var availableKeys []gpgwrapper.SecretKeyListing
+
+	allGpgKeys, err := gpg.ListSecretKeys()
 	if err != nil {
-		fmt.Printf("Failed to get fluidkeys directory")
+		fmt.Errorf("Error getting secret keys from GPG: %v", err)
```
idrysdale: Why's this `Errorf`, and above `Printf`?

idrysdale:
Cool. That looks good to me. We're still switching between 'importing', 'linking' and 'managing' keys in Fluidkeys but I think that's something for discussion for outside of this pull request!

paulfurley:
Agreed, let's follow up with a PR that unifies the language.

fluidkeys/main.go
paulfurley: because i'm a pillock

paulfurley:
I'm rebasing this now

PR #63, opened 2018-09-11
fingerprint: Fix bug with leading `0x`

paulfurley:
`strings.TrimLeft` trims *any* of `0x` so it breaks keys starting with 0

`strings.TrimPrefix` trims *the exact* `0x`, which is what we want.


idrysdale:
Ooooh, nasty. Good find 👌

PR #67, opened 2018-09-12
Print the keys managed by fluidkeys with `key list`

idrysdale:
At the moment, we print the 'Next rotation' header, but aren't checking these values for each key.

### DIFF
--- a/fluidkeys/main.go
+++ b/fluidkeys/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/fluidkeys/fluidkeys/fingerprint"
 	"github.com/fluidkeys/fluidkeys/gpgwrapper"
 	"github.com/fluidkeys/fluidkeys/humanize"
+	"github.com/fluidkeys/fluidkeys/keytableprinter"
 	"github.com/fluidkeys/fluidkeys/pgpkey"
 	"github.com/sethvargo/go-diceware/diceware"
 )
@@ -282,49 +283,8 @@ func keyList() exitCode {
 		panic(err)
 	}
 
-	emailAddressColumnWidth := len("Email address")
-
-	for _, key := range keys {
-		for _, id := range key.Identities {
-			emailAddressColumnWidth = Max(emailAddressColumnWidth, len(id.Name))
-		}
-	}
-
-	createdColumnWidth := len("31 May 2018")
-	nextRotationColumnWidth := len("Next rotation")
-
-	header := fmt.Sprintf("%-*s  ", emailAddressColumnWidth, "Email address")
-	header += fmt.Sprintf("%-*s  ", createdColumnWidth, "Created")
-	header += fmt.Sprintf("%-*s  ", nextRotationColumnWidth, "Next rotation")
-
-	fmt.Printf("%s\n", colour.LightBlue(header))
-
-	printHorizontalRule(emailAddressColumnWidth, createdColumnWidth, nextRotationColumnWidth)
-
-	for _, key := range keys {
-		firstRow := true
-		for id := range key.Identities {
-			if firstRow {
-				fmt.Printf("%-*s  ", emailAddressColumnWidth, id)
-				fmt.Printf("%-*s  ", createdColumnWidth, key.PrimaryKey.CreationTime.Format("2 Jan 2006"))
-				fmt.Printf("\n")
-				firstRow = false
-			} else {
-				fmt.Printf("%s\n", id)
-			}
-
-		}
-		printHorizontalRule(emailAddressColumnWidth, createdColumnWidth, nextRotationColumnWidth)
-	}
-
+	keytableprinter.Print(keys)
 	return 0
-}
-
-func printHorizontalRule(columnWidths ...int) {
-	for _, columnWidth := range columnWidths {
-		fmt.Printf("%s  ", strings.Repeat("─", columnWidth))
-	}
-	fmt.Printf("\n")
 }
 
 func getFluidkeysDirectory() (string, error) {
--- a/fluidkeys/main.go
+++ b/fluidkeys/main.go
@@ -61,6 +61,7 @@ func main() {
 Usage:
 	fk key create
 	fk key from-gpg
+	fk key list
 
 Options:
 	-h --help    Show this screen`, Version)
@@ -96,11 +97,13 @@ func initSubcommand(args docopt.Opts) exitCode {
 }
 
 func keySubcommand(args docopt.Opts) exitCode {
-	switch getSubcommand(args, []string{"create", "from-gpg"}) {
+	switch getSubcommand(args, []string{"create", "from-gpg", "list"}) {
 	case "create":
 		os.Exit(keyCreate())
 	case "from-gpg":
 		os.Exit(keyFromGpg())
+	case "list":
+		os.Exit(keyList())
 	}
 	panic(fmt.Errorf("keySubcommand got unexpected arguments: %v", args))
 }
@@ -264,6 +267,22 @@ func keyCreate() exitCode {
 
 	db := database.New(fluidkeysDirectory)
 	db.RecordFingerprintImportedIntoGnuPG(generateJob.pgpKey.Fingerprint())
+	return 0
+}
+
+func keyList() exitCode {
+	fluidkeysDirectory, err := getFluidkeysDirectory()
+	if err != nil {
+		fmt.Printf("Failed to get fluidkeys directory")
+	}
+	db := database.New(fluidkeysDirectory)
+
+	keys, err := loadPgpKeys(db)
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("%v\n", keys)
 	return 0
 }
 
--- a/fluidkeys/main.go
+++ b/fluidkeys/main.go
@@ -282,8 +282,49 @@ func keyList() exitCode {
 		panic(err)
 	}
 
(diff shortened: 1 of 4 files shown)


fluidkeys/main.go
```diff
@@ -267,6 +270,63 @@ func keyCreate() exitCode {
...
+		fmt.Printf("Failed to get fluidkeys directory")
+	}
+	db := database.New(fluidkeysDirectory)
+
+	keys, err := loadPgpKeys(db)
+	if err != nil {
+		panic(err)
+	}
+
+	emailAddressColumnWidth := len("Email address")
```
paulfurley: So what happens if I have an email address longer than the string `Email address` ? We talked about searching for the longest email address on the key, did that happen?

fluidkeys/main.go
```diff
@@ -267,6 +270,63 @@ func keyCreate() exitCode {
...
+
+	for _, key := range keys {
+		for _, id := range key.Identities {
+			emailAddressColumnWidth = Max(emailAddressColumnWidth, len(id.Name))
+		}
+	}
+
+	createdColumnWidth := 11
```
paulfurley: What's the origin of this magic number, is it something like `len("31 May 2018")` ? (and if so, could it just be that instead?)

fluidkeys/main.go
```diff
@@ -267,6 +270,63 @@ func keyCreate() exitCode {
 	return 0
 }
 
+func keyList() exitCode {
```
paulfurley: I think splitting out the formatting bit of this would be helpful

fluidkeys/main.go
```diff
@@ -267,6 +270,63 @@ func keyCreate() exitCode {
...
+	header := fmt.Sprintf("%-*s  ", emailAddressColumnWidth, "Email address")
+	header += fmt.Sprintf("%-*s  ", createdColumnWidth, "Created")
+	header += fmt.Sprintf("%-*s  ", nextRotationColumnWidth, "Next rotation")
+
+	fmt.Printf("%s\n", colour.LightBlue(header))
+
+	printHorizontalRule(emailAddressColumnWidth, createdColumnWidth, nextRotationColumnWidth)
+
+	for _, key := range keys {
+		firstRow := true
```
paulfurley: Some ascii-art documentation of how the table is supposed to look would be really helpful because I can't get in my head why this `firstRow` thing is necessary without thinking about the desired output

fluidkeys/main.go
```diff
@@ -267,6 +270,63 @@ func keyCreate() exitCode {
...
+			}
+
+		}
+		printHorizontalRule(emailAddressColumnWidth, createdColumnWidth, nextRotationColumnWidth)
+	}
+
+	return 0
+}
+
+func printHorizontalRule(columnWidths ...int) {
```
paulfurley: I wonder if the name is *slightly* misleading, because it's not a "horizontal rule" as I imagine from HTML, it's more a series of underlines, maybe it should be called `printHorizontalUnderlines` ?

fluidkeys/main.go
```diff
@@ -267,6 +270,63 @@ func keyCreate() exitCode {
...
+	keys, err := loadPgpKeys(db)
+	if err != nil {
+		panic(err)
+	}
+
+	emailAddressColumnWidth := len("Email address")
+
+	for _, key := range keys {
+		for _, id := range key.Identities {
+			emailAddressColumnWidth = Max(emailAddressColumnWidth, len(id.Name))
```
paulfurley: Ohh now I see `emailAddressColumnWidth` gets mutated

fluidkeys/main.go
```diff
@@ -267,6 +270,63 @@ func keyCreate() exitCode {
...
+	header += fmt.Sprintf("%-*s  ", createdColumnWidth, "Created")
+	header += fmt.Sprintf("%-*s  ", nextRotationColumnWidth, "Next rotation")
+
+	fmt.Printf("%s\n", colour.LightBlue(header))
+
+	printHorizontalRule(emailAddressColumnWidth, createdColumnWidth, nextRotationColumnWidth)
+
+	for _, key := range keys {
+		firstRow := true
+		for id := range key.Identities {
```
paulfurley: Rather than looking up `firstRow` in every iteration of the loop (when you know it'll occur exactly once, at the start), I'd suggest dropping `firstRow` and doing this:
```
for _, key := range keys {
    firstUid = key.Identities[0]
    fmt.Printf("%-*s  ", emailAddressColumnWidth, firstUid)
    fmt.Printf("%-*s  ", createdColumnWidth, key.PrimaryKey.CreationTime.Format("2 Jan 2006"))
    fmt.Printf("\n")
    for uid := range key.Identities[1:] {
        fmt.Printf("%s\n", uid)
    }
}
```

idrysdale: That would be *much* clearer, however, `Identities` is a map, not a list :/

paulfurley: Ah, this is because `.key.Identities` is a map (which also means that uids are displayed in a random order, curiously). Suggest we convert into an (ordered) slice. Function coming up...

fluidkeys/main.go
paulfurley: If I understand correctly this isn't matching the format from our design, which showed *emails*, whereas this is showing full *user ID* strings... right?

I think you might want `id.UserId.Email`

fluidkeys/main.go
paulfurley: Once #41 is merged you can do:
```go
for _, key := range keys {
    emails := key.Emails()
    if !len(emails) {
        continue
    }
    firstEmail = emails[0]
    fmt.Printf("%-*s  ", emailAddressColumnWidth, firstEmail)
    fmt.Printf("%-*s  ", createdColumnWidth, key.PrimaryKey.CreationTime.Format("2 Jan 2006"))
    fmt.Printf("\n")
    for email := range emails[1:] {
        fmt.Printf("%s\n", email)
    }
}
```

idrysdale:
I've had a go at abstracting the formating out into it's own function / file, and testing this.

Let me know what you think!

fluidkeys/main.go
```diff
@@ -267,6 +271,22 @@ func keyCreate() exitCode {
...
+		fmt.Printf("Failed to get fluidkeys directory")
+	}
+	db := database.New(fluidkeysDirectory)
+
+	keys, err := loadPgpKeys(db)
+	if err != nil {
+		panic(err)
+	}
+
+	keytableprinter.Print(keys)
```
paulfurley: simples!

keytableprinter/keytableprinter.go
```diff
@@ -0,0 +1,90 @@
...
+	fmt.Printf("%s\n", makeTable(keys))
+}
+
+func makeTable(keys []pgpkey.PgpKey) string {
+	var table string
+	table += fmt.Sprintf("%s\n", makeHeader(columnWidths(keys)))
+	table += fmt.Sprintf("%s\n", makeHorizontalUnderlines(columnWidths(keys)))
+
+	for _, key := range keys {
+		firstEmail := key.Emails(true)[0]
```
paulfurley: Remember to check that key.Emails() isn't empty, a la my (now buried) comment:
```go
    emails := key.Emails(true)
    if len(emails) {
        continue
    }
    ...
```

keytableprinter/keytableprinter_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+func TestMakeTable(t *testing.T) {
+	pgpKey, err := pgpkey.LoadFromArmoredPublicKey(exampledata.ExamplePublicKey3)
+	if err != nil {
+		t.Fatalf("Failed to load example test data: %v", err)
+	}
+
+	keys := []pgpkey.PgpKey{*pgpKey}
+
+	want := fmt.Sprintf("Email address                       Created      Next rotation\n")
```
idrysdale: I've just remembered this should be printed in `LightBlue` - will patch

keytableprinter/keytableprinter.go
paulfurley: (not really sure how to display a key that doesn't have a valid email...)

keytableprinter/keytableprinter.go
```diff
@@ -0,0 +1,90 @@
...
+		columnWidths[1] = max(columnWidths[1], len(key.PrimaryKey.CreationTime.Format("2 Jan 2006")))
+		// TODO: check on rotation date cell when implemented
+	}
+
+	return columnWidths
+}
+
+// makeHorizontalUnderlines takes an array of integers and for each integer it
+// prints a horizontal line equivalent to it's length, followed by a gutter.
+// eg [3, 2, 5] => '───  ──  ─────'
```
paulfurley: perfect comment

paulfurley:
Mmmmm
```
Email address                              Created      Next rotation
─────────────────────────────────────────  ───────────  ─────────────
foo@example.com                            7 Sep 2018 
─────────────────────────────────────────  ───────────  ─────────────
paul@fluidkeys.com                         4 Sep 2018 
─────────────────────────────────────────  ───────────  ─────────────
paul@paulfurley.com                        31 Oct 2014
─────────────────────────────────────────  ───────────  ─────────────

```

paulfurley:
I'm happy to merge once there's a [safety check on `len(key.Emails(..))`](https://github.com/fluidkeys/fluidkeys/pull/67#discussion_r217038900) :+1:

PR #69, opened 2018-09-12
Introduce Tableprinter

idrysdale:
We'll use this to print the keys back

### DIFF
--- a/tableprinter/tableprinter.go
+++ b/tableprinter/tableprinter.go
@@ -1,0 +1,86 @@
+package tableprinter
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/fluidkeys/fluidkeys/colour"
+)
+
+const gutterWidth = 2
+
+type Row struct {
+	columns []string
+}
+
+type TablePrinter struct {
+	headers Row
+	rows    []Row
+}
+
+func (tp *TablePrinter) PushRow(columnCells Row) {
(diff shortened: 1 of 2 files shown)


tableprinter/tableprinter_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/fluidkeys/fluidkeys/colour"
+)
+
+func TestAssembleTable(t *testing.T) {
+	tp := TablePrinter{
```
paulfurley: Cool amount of testing!

I was a bit surprised to see this test direct setting the struct's data, rather than calling, `PushRow` as an outside consumer would be forced to

tableprinter/tableprinter.go
```diff
@@ -0,0 +1,86 @@
...
+type TablePrinter struct {
+	headers Row
+	rows    []Row
+}
+
+func (tp *TablePrinter) PushRow(columnCells Row) {
+	tp.rows = append(tp.rows, columnCells)
+}
+
+func (tp *TablePrinter) PushHorizontalUnderlines() {
```
paulfurley: I'm not sure why this is the public interface, surely it's the `TablePrinter`'s job to determine when to push horizontal underlines, depending on the rows I've pushed?

idrysdale:
Ok, over abstracted! Closed!

PR #72, opened 2018-09-12
Use `fakeroot dpkg-deb` so files in .deb belong to root

paulfurley:


paulfurley:
Looks like `fakeroot` works!
```
$ ls -alh /bin/fk
-rwxr-xr-x 1 root root 5.0M Sep 12 15:41 /bin/fk
```

PR #77, opened 2018-09-13
Assert equal slices of strings

idrysdale:
I've done this a lot, and anticipate doing a lot more of it. Now felt as good a time as any to create a helper :)

### DIFF
--- a/assert/equality.go
+++ b/assert/equality.go
@@ -1,0 +1,15 @@
+package assert
+
+// EqualSliceOfStrings tells whether a and b contain the same elements.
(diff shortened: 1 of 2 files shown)


paulfurley:
Yeah!

How come you chose not to make it a test helper? Then it could fail with a useful error message, e.g. `expected slice[2] = "%s", got "%s"`

idrysdale:
It's true - it could be. That'd be neat!

paulfurley:
Maybe I'll sneak in a PR in a sec ;)

PR #86, opened 2018-09-14
Simplify StripAllColourCodes

paulfurley:
This way, it'll be harder for the two lists of colour codes to go out of
sync (e.g. adding a colour code to `constants.go` and not realising it
*also* must be added to `StripAllColourCodes`)


idrysdale:
Awesome 🥇

PR #91, opened 2018-09-24
Fix warnings for a key with a subkey nearing expiry

idrysdale:
I've created a really specific key here to pin the bug. Let me know if the test suite you're building makes this redundant.


exampledata/exampledata.go
```diff
@@ -155,3 +155,32 @@ D/iQVXIhUGUcj8xSftV2Iw==
 `
 
 var ExampleFingerprint3 = fingerprint.MustParse("7C18 DE4D E478 1356 8B24  3AC8 719B D63E F03B DC20")
+
```
paulfurley: LUSH

PR #93, opened 2018-09-24
Print a hint to user on `fk key list`

idrysdale:
printTopLevelHint prints a single hint to suggest the user should run`fk key rotate` if any of their keys are due or overdue for rotation.

screen shot 2018-09-24 at 15 32 16


paulfurley:
image
Was I supposed to get a hint for this state?

PR #102, opened 2018-09-25
initialize gpg, db and fluidkeys directory once

paulfurley:
Turns out there's an `init()` function which runs before main! So this is a
pretty good place to do this stuff.


idrysdale:
Great, that's all clear!

PR #104, opened 2018-09-26
PgpKey.Generate: set expiry according to (new) policy

paulfurley:
* split status package into status & policy

  status was logically doing 2 things:
  
  1. defining the "policy" e.g. when to rotate keys, define expiry etc 2.
  testing against, and enforcing the policy, e.g. "this key's expiry is too
  long"
  
  I realised this when I tried to point PgpKey.Generate at status'
  NextExpiryTime function and got a circular import[1]
  
  I think what we *actually* want is a separate "policy" package which
  defines the rules, and have status just enforce them. Then, PgpKey can ask
  the policy package for the expiry it should use.
  
  I think the policy package will also be a perfect place to define the
  ciphers, hashes, compression preferences[2]
  
  [1]:
  https://trello.com/c/KfZguoBa/182-when-we-make-a-new-key-set-its-expiry-to-the-next-rotation-time-not-60-days
  [2]:
  https://trello.com/c/5IfrDbky/185-a-newly-generated-key-doesnt-have-any-cipher-hash-or-compression-preferences

* pass creationTime into PgpKey.generateInsecure

  And add tests to check that it's actually used...
  Note that its passed down into `openpgp` inside the `Config` struct

### DIFF
--- a/pgpkey/pgpkey_test.go
+++ b/pgpkey/pgpkey_test.go
@@ -108,7 +108,7 @@ func TestFingerprintMethod(t *testing.T) {
 }
 
 func TestRevocationCertificate(t *testing.T) {
-	pgpKey, err := generateInsecure("revoke.test@example.com")
+	pgpKey, err := generateInsecure("revoke.test@example.com", time.Now())
 	if err != nil {
 		t.Fatalf("failed to generate PGP key in tests")
 	}
@@ -142,7 +142,7 @@ func TestRevocationReasonSerializeParse(t *testing.T) {
 	// Note: if this testing was in crypto/openpgp itself it could just
 	// test the unexported outSubpackets field.
 
-	pgpKey, err := generateInsecure("revoke.test@example.com")
+	pgpKey, err := generateInsecure("revoke.test@example.com", time.Now())
 	if err != nil {
 		t.Fatalf("failed to generate PGP key in tests")
 	}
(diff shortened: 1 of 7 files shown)


pgpkey/pgpkey_test.go
```diff
@@ -279,22 +281,63 @@ func TestGenerate(t *testing.T) {
 		}
 	})
 
-	t.Run("generate makes a key that by default expires in 60 days", func(*testing.T) {
-		identityName := getSingleUid(generatedKey.Identities)
-		sixtyDaysInSeconds := uint32((time.Hour * 24 * 60).Seconds())
-		actualLifetimeOfKey := generatedKey.Identities[identityName].SelfSignature.KeyLifetimeSecs
+	t.Run("PrimaryKey.CreationTime is correct", func(*testing.T) {
```
idrysdale: Good test coverage 👍

pgpkey/pgpkey_test.go
```diff
@@ -279,22 +281,63 @@ func TestGenerate(t *testing.T) {
...
+		t.Run(fmt.Sprintf("Subkeys[%d].Sig.CreationTime is correct", i), func(t *testing.T) {
+			assert.AssertEqualTimes(t, now, subkey.Sig.CreationTime)
+		})
+	}
 
-		for _, subkey := range generatedKey.Subkeys {
-			actualLifetimeOfSubkey := subkey.Sig.SigLifetimeSecs
-			if *actualLifetimeOfKey != sixtyDaysInSeconds {
-				t.Fatalf("expected KeyLifetimeSecs of Subkey to be '%v', got '%v'", sixtyDaysInSeconds, *actualLifetimeOfSubkey)
+	for name, identity := range generatedKey.Identities {
```
idrysdale: Again. So much clearer now!

pgpkey/pgpkey_test.go
```diff
@@ -279,22 +281,63 @@ func TestGenerate(t *testing.T) {
...
-		}
-	})
+
+			if expectedExpiry != *gotExpiry {
+				t.Fatalf("expected UID expiry %v, got %v", expectedExpiry, gotExpiry)
+			}
+		})
+	}
+
+	for i, subkey := range generatedKey.Subkeys {
```
idrysdale: 👍

idrysdale:
That's great.

PR #109, opened 2018-09-27
Redesign status for key list

idrysdale:
I've made the warning messages clearer, and now we nest the subkey message lines beneath the primary key messages.

We only Indent subkey warning lines only if primary warning present too.

i.e.:

```
-----------------------------------------------
Primary Key set to expire too far in the future
 └─ Subkey set to expire too far in the future
-----------------------------------------------
```

Subkey warnings remaining unindented if no primary warning present:

```
-----------------------------------------------
Subkey overdue for rotation
-----------------------------------------------
```

### DIFF
--- a/keytableprinter/formatwarnings.go
+++ b/keytableprinter/formatwarnings.go
@@ -30,42 +30,42 @@ func formatKeyWarningLines(warning status.KeyWarning) []string {
 		return []string{colour.Yellow("Missing encryption subkey")}
 
 	case status.PrimaryKeyDueForRotation:
-		return []string{colour.Yellow("Primary key due for rotation 🔄")}
+		return []string{colour.Yellow("Primary key due for rotation")}
 
 	case status.SubkeyDueForRotation:
-		return []string{colour.Yellow(" └─ Subkey due for rotation 🔄")}
+		return []string{colour.Yellow(" └─ Subkey due for rotation")}
 
 	case status.PrimaryKeyOverdueForRotation:
-		warnings := []string{colour.Red("Primary key overdue for rotation ⏰")}
+		warnings := []string{colour.Red("Primary key overdue for rotation")}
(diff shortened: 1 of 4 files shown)


paulfurley:
I'm looking at this properly now!

status/keywarning.go
```diff
@@ -90,6 +90,15 @@ func (w KeyWarning) IsAboutPrimaryKey() bool {
 	return false
 }
 
+func ContainsWarningAboutPrimaryKey(warnings []KeyWarning) bool {
```
paulfurley: Niiice

keytableprinter/formatwarnings.go
```diff
@@ -77,7 +87,7 @@ func formatKeyWarningLines(warning status.KeyWarning) []string {
 	}
 }
 
-func countdownUntilExpiry(days uint, indent uint) string {
+func countdownUntilExpiry(days uint, indent int) string {
```
paulfurley: Interesting change from `uint` to `int`, does this mean the function now supports negatives?

paulfurley: I think it's better to do `int(len(...))` in the caller rather than changing the `uint` to `int` later. Because `countdownUntilExpiry` really does only deal with positive (unsigned) integers it seems wrong to hint that it can handle a signed (possibly negative) integer by accepting `int`

paulfurley:
Awesome tests as always!

One thing that's maybe missed from the design document: it still refers to simply `Subkey` rather than `Encryption subkey`

keytableprinter/formatwarnings.go
idrysdale: Good shout! I'll address that.

paulfurley:
Perfect. will merge when travis is finished

PR #110, opened 2018-09-27
Implement key rotation

paulfurley:


fluidkeys/keyrotate.go
```diff
@@ -23,3 +215,28 @@ func printCheckboxFailure(actionText string, err error) {
...
+	}
+
+	fmt.Printf("\nRotate %s:\n\n", colour.Info(displayName(key)))
+
+	for _, action := range actions {
+		fmt.Printf("    [ ] %s\n", action)
+	}
+}
+
+func moveCursorUpLines(numLines int) {
```
idrysdale: Is this function redundant?

fluidkeys/keyrotate.go
```diff
@@ -3,8 +3,200 @@ package main
...
+		fmt.Printf("\n%s\n", colour.Green("✔ All keys look good — nothing to do."))
+		return 0 // success! nothing to do
+	}
+
+	if !runImportBackIntoGnupg(keysModifiedSuccessfully, passwords) {
+		numErrorsEncountered += 1
+	}
+
+	if numErrorsEncountered > 0 {
+		message := fmt.Sprintf("%d error(s) while running rotate.", numErrorsEncountered)
```
idrysdale: Could you use `humanize.Pluralize` here to make it a touch nicer in the user interface?

paulfurley: Ah great plan

fluidkeys/keyrotate.go
paulfurley: It *was*, but I've added commit https://github.com/fluidkeys/fluidkeys/pull/110/commits/429546895c3cc7b42ee6f50c16b0f09962971d17 which uses it instead of "\z"

PR #111, opened 2018-09-27
gpg:ExportPrivateKey: return a BadPasswordError

paulfurley:


idrysdale:
Nice one. Good way of bubbling this up!

PR #116, opened 2018-10-03
fix broken TestGetEncryptionSubkeyWarnings

paulfurley:
Rather than using the ExamplePublicKey4, whcih relied on *real-world* now
rather than a test now, use ExamplePrivateKey2 and tweak its subkey and
primary key expiration times in the test.

* rename & export `PgpKey.updateSubPgpKeyExpiryToNow` as `PgpKey.UpdateSubPgpKeyValidUntil`
* pgpkey:EncryptionSubkey: take optional `now` arg
* delete ExamplePublicKey4

  - it has a short expiry date, so it has now expired (the test was written
to depend on *real* now rather than passed-in `now`)
  - the private key is not committed, so I'm unable to edit the key :(

### DIFF
--- a/exampledata/exampledata.go
+++ b/exampledata/exampledata.go
@@ -155,32 +155,3 @@ D/iQVXIhUGUcj8xSftV2Iw==
 `
 
 var ExampleFingerprint3 = fingerprint.MustParse("7C18 DE4D E478 1356 8B24  3AC8 719B D63E F03B DC20")
-
-const ExamplePublicKey4 = `
------BEGIN PGP PUBLIC KEY BLOCK-----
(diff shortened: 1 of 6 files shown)


paulfurley:
@idrysdale thinking about this more last night (after merging), actually there was nothing wrong with using the example key regardless of its expiry date - *except* that we hadn't yet implemented a `now` parameter on `PgpKey.EncryptionSubkey()`. That meant that `getSubkeyWarnings` didn't work as expected in tests.

So in retrospect, just doing 182f395be349dec46ef75ee0907fbbde8feba37c would have fixed this.

Given that the example key didn't have its private key part I think it's still right to delete it, since it's impossible to use for certain types of tests, and can't be edited.

PR #122, opened 2018-10-08
Improve the public interface to Colour

idrysdale:
I was getting in a bit of a muddle with whether we refer to colours by
actual colour or abstract style and have opted to export only abstract
style functions like 'Info', 'Warning', or 'Success'. These are then
defined within Colour itself.


paulfurley:
I've just rebased this

paulfurley:
Love it

PR #123, opened 2018-10-08
Make human friendly messages for all warnings in table

idrysdale:
Note, this PR depends on https://github.com/fluidkeys/fluidkeys/pull/122

It's made sense to move these user friendly messages onto `KeyWarning.String()` now, much like we have done with the `KeyAction` interface.

I've also dropped all code referring to any indentation, or └─, which has made things much cleaner in the code.


keytableprinter/keytableprinter.go
```diff
@@ -32,6 +32,7 @@ func Print(keys []pgpkey.PgpKey) {
 	}
 
 	fmt.Print(makePrimaryInstruction(keys))
+	fmt.Println()
```
paulfurley: does this happen even if `makePrimaryInstruction` returns `""` ?

PR #127, opened 2018-10-10
add logic for new KeyWarning -> KeyAction

paulfurley:
Note that this still doesn't *implement* the actions but it does the logic on
mapping from a bunch of KeyWarnings to KeyActions

### DIFF
--- a/status/actions.go
+++ b/status/actions.go
@@ -1,6 +1,8 @@
 package status
 
 import (
+	"fmt"
+	"sort"
 	"time"
 
 	"github.com/fluidkeys/fluidkeys/policy"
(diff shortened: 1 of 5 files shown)


status/actions.go
```diff
@@ -9,14 +11,42 @@ import (
 // MakeActionsFromWarnings returns a list of actions that can be performed on
 // the key to fix the warning.
 // Call `KeyAction.Enact(key)` to actually carry out the action.
-func MakeActionsFromWarnings(warnings []KeyWarning) []KeyAction {
-	now := time.Now()
+func MakeActionsFromWarnings(warnings []KeyWarning, args ...time.Time) []KeyAction {
```
idrysdale: It's a neat trick to use `args` for time.Now(), but think we should specifically call it `now` and not be optional. Bit more obvs.

status/actions.go
```diff
@@ -9,14 +11,42 @@ import (
...
 
 	var actions []KeyAction
 	for _, warning := range warnings {
 		actions = append(actions, makeActionsFromSingleWarning(warning, now)...)
 	}
-	return actions
+	return deduplicateAndOrder(actions)
+}
+
+func deduplicateAndOrder(actions []KeyAction) []KeyAction {
```
idrysdale: I like the sound of this.

paulfurley: ;)

status/status.go
```diff
@@ -241,23 +241,23 @@ func joinHashNames(hashes []uint8) string {
 	for _, hashByte := range hashes {
 		hashNames = append(hashNames, hash.Name(hashByte))
 	}
-	return strings.Join(hashNames, " ")
+	return strings.Join(hashNames, ", ")
```
idrysdale: Agreed, much nicer 👍

idrysdale:
This makes me happy.

PR #134, opened 2018-10-10
Start storing backups in date-stamped subdirectories

idrysdale:
This subdirectory is created from the date at which the backup is created (as opposed to the key's slug, which is the date at which the key was created).

We also now print the full path of the backup directory, rather than just the fluidkeys directory.


paulfurley:
```
fluidkeys/main.go:250:45: syntax error: unexpected directory, expecting comma or )
```

paulfurley:
this is brilliant!

PR #136, opened 2018-10-11
fix UpdateSubkeyValidUntil — pass in `now` arg

paulfurley:
(rebase after merging #135)

previously `validUntil` was being used as the signature `CreationTime`.

That's a trap: if you updated the validity to the years in the future, the
future-dated signature would override any new signatures trying to shorten
the subkey expiry (because the future-dated signature would always be the
"newest")

### DIFF
--- a/pgpkey/pgpkey.go
+++ b/pgpkey/pgpkey.go
@@ -492,7 +492,9 @@ func (key *PgpKey) createNewEncryptionSubkey(validUntil time.Time, now time.Time
 
 // ExpireSubkey prevents the given subkey from being usable.
 func (key *PgpKey) ExpireSubkey(subkeyId uint64) error {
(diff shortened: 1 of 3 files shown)


idrysdale:
Rebased!

pgpkey/pgpkey.go
```diff
@@ -523,7 +525,7 @@ func (key *PgpKey) UpdateSubkeyValidUntil(subkeyId uint64, validUntil time.Time)
 
 	subkey.Sig.SigType = packet.SigTypeSubkeyBinding
 	subkey.Sig.Hash = config.Hash()
-	subkey.Sig.CreationTime = validUntil // essential that this sig is the most recent
+	subkey.Sig.CreationTime = now // essential that this sig is the most recent
```
idrysdale: Good spot!

PR #139, opened 2018-10-11
Backup all modified keys on `key rotate`

idrysdale:
This ensures the user will always have a backup to disk of their latest
    keys. Note: if a user imports a key from GPG into Fluidkeys and
    Fluidkeys subsequently makes any changes to this key, it'll then go on
    to back this up to disk, too. It does this automatically.

### DIFF
--- a/fluidkeys/keyrotate.go
+++ b/fluidkeys/keyrotate.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"fmt"
+	"path/filepath"
 	"time"
 
+	"github.com/fluidkeys/fluidkeys/backupzip"
 	"github.com/fluidkeys/fluidkeys/colour"
 	"github.com/fluidkeys/fluidkeys/fingerprint"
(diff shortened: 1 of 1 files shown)


paulfurley:
Tea, then I'm reviewing this!

fluidkeys/keyrotate.go
```diff
@@ -206,6 +222,7 @@ func printImportBackIntoGnupg(keys []*pgpkey.PgpKey) {
 
 	for _, key := range keys {
 		fmt.Printf("     [ ] Import %s back into GnuPG\n", displayName(key))
+		fmt.Printf("     [ ] Backup %s\n", displayName(key))
```
paulfurley: I think it would be easier on the eye if it showed all the "import back into..." followed by "backup"... rather than alternating 
image

paulfurley:
@idrysdale I'm gonna merge and link back to my review comments in the design crit card :)

PR #140, opened 2018-10-11
Use proper key sizes in PgpKey

paulfurley:
* re-write PgpKey.Generate based on openpgp.NewEntity

	the reason for abandoning openpgp.NewEntity is that it is unable to make a
	primary key with a different key size than the subkey.

	further, we have a lot of the functionality ourselves already
	(CreateNewEncryptionSubkey) so it's more consistent to just use that.

	note that we never generate small 1024-bit keys for testing any more,
	instead we use the mockRandom method. this allows us to really test that
	the code is working with the correct key sizes.

* add policy RSA key lengths for primary/subkey and tests

* add mockRandom to pgpkey_tests to allow for faster testing

	this uses the insecure math.random pseudorandom number generator instead of
	the (slow) crypto/rand

* CreateNewEncryptionSubkey: add random io.Reader so that 

* pgpkey_tests: load example key over generating key

	it's *extremely* expensive to generate a 4096-bit key, even with a fast
	insecure random number generator.

	instead, load these keys from exampledata

* delete no-longer-used Config{} struct


status/keyactions.go
```diff
@@ -44,7 +44,7 @@ type CreateNewEncryptionSubkey struct {
 }
 
 func (a CreateNewEncryptionSubkey) Enact(key *pgpkey.PgpKey, now time.Time) error {
-	return key.CreateNewEncryptionSubkey(a.ValidUntil, now)
+	return key.CreateNewEncryptionSubkey(a.ValidUntil, now, nil)
```
idrysdale: That's interesting. So passing nil to `packet.Config{Rand}` means it uses a cryptographically random number?

pgpkey/pgpkey.go
```diff
@@ -40,8 +35,11 @@ func (e *IncorrectPassword) Error() string {
 	return fmt.Sprintf("incorrect password: %s", e.decryptErrorMessage)
 }
 
-func Generate(email string, now time.Time) (*PgpKey, error) {
-	return generateKeyOfSize(email, RsaSizeSecureKeyBits, now)
+func Generate(email string, now time.Time, random io.Reader) (*PgpKey, error) {
+	if random == nil {
+		random = cryptorand.Reader
```
idrysdale: Ah - here it is!

paulfurley: Yeah indeed. That could do with an update to the doc comment

status/keyactions.go
paulfurley: I've updated the docs

PR #149, opened 2018-10-15
read passwords from FLUIDKEYS_PASSWORDS_TOML_FILE

paulfurley:
(merge after merging #148)

workaround for this bug:

https://trello.com/c/iwwtzNbt/223-on-macos-keychain-access-prompts-every-time-fluidkeys-tries-to-access-a-password

if specified in the environment, read passwords from the given TOML file
and attempt to use them before looking in the keyring.

this enables people to start using automated (cron) key rotation if they
choose to, before we fix the issue with Keychain Access


idrysdale:
Rebasing...

PR #153, opened 2018-10-16
Prompt to automatically rotate successfully rotated keys

idrysdale:
Not that while `SetStorePassword` will get set in the config at this point,
the user hasn't been given a further opportunity to input their password
so it won't be stored in the keyring.


idrysdale:
Merge after #154

PR #154, opened 2018-10-16
Make promptYesOrNo take defaultInput rather than defaultResult

idrysdale:
* Print the options depending on defaultValue passed to `promptYesOrNo`

### DIFF
--- a/fluidkeys/main.go
+++ b/fluidkeys/main.go
@@ -353,7 +353,7 @@ func promptForKeyToImportFromGpg(secretKeyListings []gpgwrapper.SecretKeyListing
 	var selectedKey int
 	if len(secretKeyListings) == 1 {
 		onlyKey := secretKeyListings[0]
-		if promptYesOrNo("Import key? [Y/n] ", true) {
+		if promptYesOrNo("Import key? [Y/n] ", "y") {
 			return &onlyKey
 		} else {
 			return nil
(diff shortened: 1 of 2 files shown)


fluidkeys/main.go
```diff
@@ -378,12 +378,19 @@ func promptForKeyToImportFromGpg(secretKeyListings []gpgwrapper.SecretKeyListing
 	}
 }
 
-func promptYesOrNo(message string, defaultResult bool) bool {
+func promptYesOrNo(message string, defaultInput string) bool {
 	for {
-		input := promptForInput(message)
+		var options string
+		switch strings.ToLower(defaultInput) {
+		case "y":
```
paulfurley: clever!!

fluidkeys/main.go
```diff
@@ -394,23 +401,28 @@ func promptYesOrNo(message string, defaultResult bool) bool {
...
+	if input != "" {
+		return input
+	} else {
+		return defaultInput
+	}
 }
 
-func promptForInput(prompt string) string {
-	return promptForInputWithPipes(prompt, bufio.NewReader(os.Stdin))
+func promptForInput(prompt string, defaultInput string) string {
```
paulfurley: I'm not sure the value of adding a `defaultInput` to this generic `promptForInput`function? 

AFAICT there are no callers `promptForInput` that want to specify a default input, so they all end up having to add `""` to their argument list, which is weird

idrysdale: Agreed, I'll move it down a level into `promptForInputWithPipes` and call that directly from `promptYesOrNo`

paulfurley: I think that' still overcomplicating things. Why not do this logic in `promptYesOrNo` ? that's the only place that the default makes sense

PR #158, opened 2018-10-17
add fluidkeys to user's crontab

paulfurley:
* add scheduler.Enable, Disable
* enable or disable cron (scheduler) from init
* add Config.RunFromCron()
* add Config.Filename()...
* add ui.go with printInfo ... this is a small step to factoring out some of the UI components

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,18 @@ func (c *Config) GetFilename() string {
 	return c.filename
 }
 
+func (c *Config) RunFromCron() bool {
+	if !c.parsedMetadata.IsDefined("run_from_cron") {
+		c.parsedConfig.RunFromCron = defaultRunFromCron
+		err := c.save()
+		if err != nil {
+			panic(err)
+		}
+	}
+
+	return c.parsedConfig.RunFromCron
+}
+
 // ShouldStorePasswordForKey returns whether the given key's password should
 // be stored in the system keyring when successfully entered (avoiding future
 // password prompts).
@@ -191,7 +203,8 @@ const (
 )
 
 type tomlConfig struct {
-	PgpKeys map[string]key `toml:"pgpkeys"`
+	RunFromCron bool           `toml:"run_from_cron"`
+	PgpKeys     map[string]key `toml:"pgpkeys"`
 }
 
 type key struct {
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -57,6 +57,18 @@ func (c *Config) GetFilename() string {
 	return c.filename
 }
 
+func (c *Config) RunFromCron() bool {
```
idrysdale: Ah hah, cunning. So it looks like we're going to record in our `config` if we expect to be running from `cron` or not...

config/config.go
```diff
@@ -213,14 +213,29 @@ type key struct {
...
+#     # store_password tells Fluidkeys to use the system keychain to store
+#     # the password for this key and look for it before prompting.
 #     store_password = true
+#
+#     # rotate_automatically specifies that key rotation tasks should be
+#     # carried out without prompting when running 'fk key rotate --automatic'
+#     # store_password must also be true to rotate keys automatically.
 #     rotate_automatically = true
+#
+# THIS FILE IS OVERWRITTEN BY FLUIDKEYS.
```
idrysdale: Ah, that's sensible. Good documentation 💪

scheduler/scheduler.go
```diff
@@ -0,0 +1,124 @@
...
+	return removeCrontabLines(crontab) + cronLines
+}
+
+func removeCrontabLines(crontab string) string {
+	return strings.Replace(crontab, cronLines, "", -1)
+}
+
+const crontab string = "crontab"
+const cronLines string = `
+# Fluidkeys added the following line. To disable, edit ~/.config/fluidkeys/config.toml
```
idrysdale: IIRC Fluidkeys directory can be different. Is it possible to dynamically create this line for them to specify their actual directory?

paulfurley: Interesting point. The problem is that we want these two lines to stay *exaclty* the same between runs to make it trivial to parse them in an existing crontab. If they can change (including if we change them in the future) then we can't detect them anymore.

Maybe the thing to do is just say "To disable, edit your Fluidkeys configuration file"

idrysdale: 👍

PR #161, opened 2018-10-18
specify full path in crontab

paulfurley:


paulfurley:
@idrysdale after you merge this if you get duplicate entries (see `crontab -l`) then do `crontab -r` to delete your crontab

PR #168, opened 2018-10-19
Timestamp all file backups

idrysdale:
Rather than just put backup zips in subdirectories of dates, we now put the
time stamp on the end of each file.

i.e: `/backups/2018-10-19/gpghome-110713.tgz`

### DIFF
--- a/archiver/archiver.go
+++ b/archiver/archiver.go
@@ -6,7 +6,22 @@ import (
 	"time"
 )
 
-func DateStampedDirectory(fluidkeysDir string, now time.Time) string {
+// MakeFilePath takes a filename, extension, directory and time and constructs
+// a filepath string formated like:
+//   directory/2016-08-23/filename-180500.ext
+func MakeFilePath(filename string, extension string, directory string, now time.Time) string {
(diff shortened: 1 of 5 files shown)


archiver/archiver.go
```diff
@@ -6,7 +6,22 @@ import (
...
+//   directory/2016-08-23/filename-180500.ext
+func MakeFilePath(filename string, extension string, directory string, now time.Time) string {
+	return filepath.Join(
+		dateStampedDirectory(directory, now),
+		appendTimeStampToFilename(filename, extension, now),
+	)
+}
+
+func appendTimeStampToFilename(filename string, extension string, now time.Time) string {
+	timestamp := now.Format("150405")
```
paulfurley: Cool!

Though i'd suggest it would make more sense to include the date (again) because if the backup file were ever moved (like you might if you ever needed to *use* it), it's a meaningless filename.

Adding hyphens would make it more obvious too, maybe now.Format("2006-01-02T15-04-05") (*almost* ISO 8601 but avoiding colons in the time)

PR #169, opened 2018-10-21
Cron output

paulfurley:
@idrysdale I opened this as a PR so I have an interface for leaving comments!

### DIFF
--- a/fluidkeys/keyrotate.go
+++ b/fluidkeys/keyrotate.go
@@ -16,7 +16,7 @@ import (
 	"github.com/fluidkeys/fluidkeys/status"
 )
 
-func keyRotate(dryRun bool, automatic bool) exitCode {
+func keyRotate(dryRun bool, automatic bool, cronOutput bool) exitCode {
 	keys, err := loadPgpKeys()
 	if err != nil {
 		panic(err)
@@ -38,7 +38,11 @@ func keyRotate(dryRun bool, automatic bool) exitCode {
 		}
 
 		var outputter outputterInterface
-		outputter = &terminalOutputter{}
+		if cronOutput {
+			outputter = &cronOutputter{}
+		} else {
+			outputter = &terminalOutputter{}
+		}
 		return runKeyRotate(keys, yesNoPrompter, passwordPrompter, outputter)
 	}
 }
--- a/fluidkeys/keyrotate.go
+++ b/fluidkeys/keyrotate.go
@@ -11,6 +11,7 @@ import (
 	"github.com/fluidkeys/fluidkeys/colour"
 	"github.com/fluidkeys/fluidkeys/fingerprint"
 	"github.com/fluidkeys/fluidkeys/humanize"
+	"github.com/fluidkeys/fluidkeys/out"
 	"github.com/fluidkeys/fluidkeys/pgpkey"
 	"github.com/fluidkeys/fluidkeys/scheduler"
 	"github.com/fluidkeys/fluidkeys/status"
@@ -22,7 +23,7 @@ func keyRotate(dryRun bool, automatic bool, cronOutput bool) exitCode {
 		panic(err)
 	}
 
-	fmt.Printf("\n")
+	out.Print("\n")
 	if dryRun {
 		return runKeyRotateDryRun(keys)
 	} else {
@@ -37,13 +38,10 @@ func keyRotate(dryRun bool, automatic bool, cronOutput bool) exitCode {
 			passwordPrompter = &alwaysFailPasswordPrompter{}
 		}
 
-		var outputter outputterInterface
 		if cronOutput {
-			outputter = &cronOutputter{}
-		} else {
-			outputter = &terminalOutputter{}
+			out.Outputter = &out.CronOutputter{}
 		}
-		return runKeyRotate(keys, yesNoPrompter, passwordPrompter, outputter)
+		return runKeyRotate(keys, yesNoPrompter, passwordPrompter)
 	}
 }
 
@@ -54,25 +52,24 @@ var (
 )
 
 func runKeyRotateDryRun(keys []pgpkey.PgpKey) exitCode {
-	outputter := &terminalOutputter{}
 	keyTasks := makeKeyTasks(keys)
 
 	if len(keyTasks) == 0 {
-		outputter.print(nothingToDo)
+		out.Print(nothingToDo)
 		return 0 // success! nothing to do
 	}
 
-	outputter.print(reviewTheseActions)
+	out.Print(reviewTheseActions)
 
 	for i := range keyTasks {
 		var keyTask *keyTask = keyTasks[i]
 		keyTask.actions = addImportExportActions(keyTask.actions, nil)
-		printKeyWarningsAndActions(*keyTask, outputter)
+		printKeyWarningsAndActions(*keyTask)
 	}
 
-	outputter.print("To start run\n")
-	outputter.print(" >   " + colour.CommandLineCode("fk key rotate") + "\n\n")
-	outputter.print("You’ll be asked at each stage to confirm before making any changes.\n\n")
+	out.Print("To start run\n")
+	out.Print(" >   " + colour.CommandLineCode("fk key rotate") + "\n\n")
+	out.Print("You’ll be asked at each stage to confirm before making any changes.\n\n")
 	return 0
 }
 
@@ -94,12 +91,12 @@ const (
 )
 
 type promptYesNoInterface interface {
-	promptYesNo(message string, defaultResponse string, key *pgpkey.PgpKey, outputter outputterInterface) bool
+	promptYesNo(message string, defaultResponse string, key *pgpkey.PgpKey) bool
 }
 
 type interactiveYesNoPrompter struct{}
 
-func (iP *interactiveYesNoPrompter) promptYesNo(message string, defaultInput string, key *pgpkey.PgpKey, outputter outputterInterface) bool {
+func (iP *interactiveYesNoPrompter) promptYesNo(message string, defaultInput string, key *pgpkey.PgpKey) bool {
 	var options string
 	switch strings.ToLower(defaultInput) {
 	case "y":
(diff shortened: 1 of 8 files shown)


paulfurley:
Been thinking about this on and off over the weekend, got some random thoughts:

* any printing that happens *before* we start using the cron outputter (e.g. things that happen in the `init` functions) won't be intercepted. I think we've got 2 options here:
    1. don't do any output in `init` functions (except for if we're about to panic)
    2. move the init functions into `main`, right after parsing `--cron-output`

* maybe this is the time to lay the groundwork for sending *all* output through our own outputter object? e..g. have a *global* formatter interface which is set up in `init`.  The `--cron-output` flag could switch out the default outputter with the cron outputter. Then key rotate can just use `Out.Print(...)` etc rather than passing an interface around

[more comments later]

fluidkeys/keyrotate.go
```diff
@@ -152,17 +160,42 @@ func (p *alwaysFailPasswordPrompter) promptForPassword(key *pgpkey.PgpKey) (stri
...
+
+type cronOutputter struct {
+	buffer string
+}
+
+func (o *cronOutputter) print(message string) {
+	o.buffer += message
+}
+
+func (o *cronOutputter) printBuffer() {
```
paulfurley: I'm not sure what `printBuffer` means... could be:
* print *into* buffer (not to console)
* print *the* buffer (dump whatever's in the buffer to console)

idrysdale: Yep, good shout. Renaming to `printTheBuffer`

idrysdale: Yep, good shout. Renaming to `printTheBuffer`

> Been thinking about this on and off over the weekend, got some random thoughts:
> 
> * any printing that happens _before_ we start using the cron outputter (e.g. things that happen in the `init` functions) won't be intercepted. I think we've got 2 options here:
>   
>   1. don't do any output in `init` functions (except for if we're about to panic)
>   2. move the init functions into `main`, right after parsing `--cron-output`
> * maybe this is the time to lay the groundwork for sending _all_ output through our own outputter object? e..g. have a _global_ formatter object (interface) which is set up in `init`.
>   
>   * The `--cron-output` flag could switch out the default outputter with the cron outputter.
>   * Key rotate can just use `Out.Print(...)` on the global pointer rather than passing an interface around

Great suggestion. Have implemented. It's nice only having fmt.Print() in one place now!

>   * Q: do packages have access to a global defined in the `main` package?

Not as far as I'm aware... Therefore I've had a go at creating a new package `out` that holds the `Outputter` and has two functions `Print` and `PrintTheBuffer`. I'm not sure this is quite right, but it's working.

idrysdale: Yep, good shout. Renaming to `printTheBuffer`

> Been thinking about this on and off over the weekend, got some random thoughts:
> 
> * any printing that happens _before_ we start using the cron outputter (e.g. things that happen in the `init` functions) won't be intercepted. I think we've got 2 options here:
>   
>   1. don't do any output in `init` functions (except for if we're about to panic)
>   2. move the init functions into `main`, right after parsing `--cron-output`
> * maybe this is the time to lay the groundwork for sending _all_ output through our own outputter object? e..g. have a _global_ formatter object (interface) which is set up in `init`.
>   
>   * The `--cron-output` flag could switch out the default outputter with the cron outputter.
>   * Key rotate can just use `Out.Print(...)` on the global pointer rather than passing an interface around

Great suggestion. Have implemented. It's nice only having fmt.Print() in one place now!

>   * Q: do packages have access to a global defined in the `main` package?

Not as far as I'm aware... Therefore I've had a go at creating a new package `out` that holds the `Outputter` and has two functions `Print` and `PrintTheBuffer`. I'm not sure this is quite right, but it's working.

idrysdale:
Yep, good shout. Renaming to `printTheBuffer`

> Been thinking about this on and off over the weekend, got some random thoughts:
> 
> * any printing that happens _before_ we start using the cron outputter (e.g. things that happen in the `init` functions) won't be intercepted. I think we've got 2 options here:
>   
>   1. don't do any output in `init` functions (except for if we're about to panic)
>   2. move the init functions into `main`, right after parsing `--cron-output`
> * maybe this is the time to lay the groundwork for sending _all_ output through our own outputter object? e..g. have a _global_ formatter object (interface) which is set up in `init`.
>   
>   * The `--cron-output` flag could switch out the default outputter with the cron outputter.
>   * Key rotate can just use `Out.Print(...)` on the global pointer rather than passing an interface around

Great suggestion. Have implemented. It's nice only having fmt.Print() in one place now!

>   * Q: do packages have access to a global defined in the `main` package?

Not as far as I'm aware... Therefore I've had a go at creating a new package `out` that holds the `Outputter` and has two functions `Print` and `PrintTheBuffer`. I'm not sure this is quite right, but it's working.

idrysdale:
Yep, good shout. Renaming to `printTheBuffer`

> Been thinking about this on and off over the weekend, got some random thoughts:
> 
> * any printing that happens _before_ we start using the cron outputter (e.g. things that happen in the `init` functions) won't be intercepted. I think we've got 2 options here:
>   
>   1. don't do any output in `init` functions (except for if we're about to panic)
>   2. move the init functions into `main`, right after parsing `--cron-output`
> * maybe this is the time to lay the groundwork for sending _all_ output through our own outputter object? e..g. have a _global_ formatter object (interface) which is set up in `init`.
>   
>   * The `--cron-output` flag could switch out the default outputter with the cron outputter.
>   * Key rotate can just use `Out.Print(...)` on the global pointer rather than passing an interface around

Great suggestion. Have implemented. It's nice only having fmt.Print() in one place now!

>   * Q: do packages have access to a global defined in the `main` package?

Not as far as I'm aware... Therefore I've had a go at creating a new package `out` that holds the `Outputter` and has two functions `Print` and `PrintTheBuffer`. I'm not sure this is quite right, but it's working.

idrysdale:
Thanks for this review. Super helpful.

I've:

1. Renamed `printBuffer` > `printTheBuffer`
2. Created a global `out` package, holding the `OutputterInterface`

>   * Q: do packages have access to a global defined in the `main` package?

They didn't. This didn't appear to be a problem at first, as we _mostly_ print from `main`, but then I found a couple of places where this wasn't the case, for example `keytableprinter`.

To give that package access to the interface, I had to factor it out into it's own package. I called this `out`. I'm not sure this is quite right, but it's working. The other approach would have been to keep the `out` global variable on `main` and pass it into to the various other packages that then will output, but I had trouble defining the type they were expecting (i.e. `OutputterInterface` wasn't defined)

Maybe best to talk this through when you're back tomorrow :)

idrysdale:
Thanks for this review. Super helpful.

I've:

1. Renamed `printBuffer` > `printTheBuffer`
2. Created a global `out` package, holding the `OutputterInterface`

>   * Q: do packages have access to a global defined in the `main` package?

They didn't. This didn't appear to be a problem at first, as we _mostly_ print from `main`, but then I found a couple of places where this wasn't the case, for example `keytableprinter`.

To give that package access to the interface, I had to factor it out into it's own package. I called this `out`. I'm not sure this is quite right, but it's working. The other approach would have been to keep the `out` global variable on `main` and pass it into to the various other packages that then will output, but I had trouble defining the type they were expecting (i.e. `OutputterInterface` wasn't defined)

Maybe best to talk this through when you're back tomorrow :)

idrysdale:
Thanks for your review, it was super helpful.

I've:

1. Renamed `printBuffer` > `printTheBuffer`
2. Created a global `out` package, holding the `OutputterInterface`

>   * Q: do packages have access to a global defined in the `main` package?

They didn't. This didn't appear to be a problem at first, as we _mostly_ print from `main`, but then I found a couple of places where this wasn't the case, for example `keytableprinter`.

To give that package access to the interface, I had to factor it out into it's own package. I called this `out`. I'm not sure this is quite right, but it's working. The other approach would have been to keep the `out` global variable on `main` and pass it into to the various other packages that then will output, but I had trouble defining the type they were expecting (i.e. `OutputterInterface` wasn't defined)

Maybe best to talk this through when you're back tomorrow :)

idrysdale:
Thanks for your review, it was super helpful.

I've:

1. Renamed `printBuffer` > `printTheBuffer`
2. Created a global `out` package, holding the `OutputterInterface`

>   * Q: do packages have access to a global defined in the `main` package?

They didn't. This didn't appear to be a problem at first, as we _mostly_ print from `main`, but then I found a couple of places where this wasn't the case, for example `keytableprinter`.

To give that package access to the interface, I had to factor it out into it's own package. I called this `out`. I'm not sure this is quite right, but it's working. The other approach would have been to keep the `out` global variable on `main` and pass it into to the various other packages that then will output, but I had trouble defining the type they were expecting (i.e. `OutputterInterface` wasn't defined)

Maybe best to talk this through when you're back tomorrow :)

idrysdale:
Thanks for your review, it was super helpful.

I've:

1. Renamed `printBuffer` > `printTheBuffer`
2. Created a global `out` package, holding the `OutputterInterface`

>   * Q: do packages have access to a global defined in the `main` package?

They didn't. This didn't appear to be a problem at first, as we _mostly_ print from `main`, but then I found a couple of places where this wasn't the case, for example `keytableprinter`.

To give that package access to the interface, I had to factor it out into it's own package. I called this `out`. I'm not sure this is quite right, but it's working. The other approach would have been to keep the `out` global variable on `main` and pass it into to the various other packages that then will output, but I had trouble defining the type they were expecting (i.e. `OutputterInterface` wasn't defined)

Maybe best to talk this through when you're back tomorrow :)

idrysdale:
Thanks for your review, it was super helpful.

I've:

1. Renamed `printBuffer` > `printTheBuffer`
2. Created a global `out` package, holding the `OutputterInterface`

>   * Q: do packages have access to a global defined in the `main` package?

They didn't. This didn't appear to be a problem at first, as we _mostly_ print from `main`, but then I found a couple of places where this wasn't the case, for example `keytableprinter`.

To give that package access to the interface, I had to factor it out into it's own package. I called this `out`. I'm not sure this is quite right, but it's working. The other approach would have been to keep the `out` global variable on `main` and pass it into to the various other packages that then will output, but I had trouble defining the type they were expecting (i.e. `OutputterInterface` wasn't defined)

Maybe best to talk this through when you're back tomorrow :)

PR #194, opened 2018-10-27
don't output stray newline in cron mode

paulfurley:
... as it causes cron the think the job failed, and send emails


paulfurley:
@idrysdale forgive me for self merging (and reprimand me tomorrow), but I'm keen to fix this simple-but-annoying issue for @tomtaylor and @frabcus

PR #204, opened 2018-10-30
move code out of fluidkeys/main.go

paulfurley:
* move init functions into fluidkeys/init.go
* un-factor-out constant strings in keycreate
* move PromptWhichKeyFromGPG into fluidkeys/keyfromgpg
* move DicewarePassword / generatePgpKeyResult - these should've been moved into fluidkeys/keycreate.go before
* move ensureCrontabStateMatchesConfig below main


paulfurley:
travis *has* tested both, i don't know why it's not updating

PR #206, opened 2018-10-30
Tell the user which keyring their password is stored in

idrysdale:
This involved updating `fluidkeys/keyring` to return the Backendtype.


idrysdale:
Right! Nice one!

PR #207, opened 2018-10-31
Log user input on promptForInputWithPipes

paulfurley:
This does, however, mean we're logging the nth word of their password
we ask them for.

Thoughts?

### DIFF
--- a/fluidkeys/keycreate.go
+++ b/fluidkeys/keycreate.go
@@ -43,7 +43,7 @@ func keyCreate() exitCode {
 			"later import it from your backup.\n\n" +
 			"Alternatively, quit now [ctrl-c], install GPG then " +
 			"run FluidKeys again.\n\n")
-		promptForInput("Press enter to continue. ")
+		promptForInput("Press enter to continue. ", true)
 	}
 	out.Print("\n")
 	email := promptForEmail()
@@ -105,7 +105,7 @@ func generatePgpKey(email string, channel chan generatePgpKeyResult) {
 
 func promptForEmail() string {
 	out.Print(PromptEmail + "\n")
-	return promptForInput("[email] : ")
+	return promptForInput("[email] : ", true)
 }
 
 func generatePassword(numberOfWords int, separator string) DicewarePassword {
@@ -120,7 +120,7 @@ func displayPassword(password DicewarePassword) {
 	out.Print("If you use a password manager, save it there now.\n\n")
 	out.Print(colour.Warning("Store this safely, otherwise you won’t be able to use your key\n\n"))
 
-	promptForInput("Press enter when you've stored it safely. ")
+	promptForInput("Press enter when you've stored it safely. ", true)
 }
 
 func userConfirmedRandomWord(password DicewarePassword) bool {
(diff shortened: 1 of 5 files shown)


paulfurley:
The outer `promptForInput` function is only called in a few places, so I suggest we should be explicit and add a `logToFile` bool on that function. It'll mean updating all the callers, but I think it's good for future prompts too that we have to make a decision whether to log the input to file or not (e.g. when we implement secret snippets or something)
```
fluidkeys/keycreate.go:28:		promptForInput("Press enter to continue. ")
fluidkeys/keycreate.go:90:	return promptForInput("[email] : ")
fluidkeys/keycreate.go:105:	promptForInput("Press enter when you've stored it safely. ")
fluidkeys/keycreate.go:116:	givenWord := promptForInput("[" + wordOrdinal + " word] : ")
fluidkeys/keymaintain.go:117:		input := promptForInput(messageWithOptions)
fluidkeys/main.go:272:func promptForInput(prompt string) string {
fluidkeys/keyfromgpg.go:125:			input := promptForInput(fmt.Sprintf(PromptWhichKeyFromGPG + " " + rangePrompt + " "))
```

idrysdale:
Good idea, I'll pick that up now.

paulfurley:
Some quirks:

1. The "press enter" prompts aren't very useful to log, since they produce lines like this:

```
2018/11/02 13:46:00 Press enter when you've stored it safely. 
2018/11/02 13:46:11 > 
2018/11/02 13:46:11 [H[2J
```

Suggestion: only output lines that don't evaluate to `""` (probably after calling `strings.Trim("<something>")`)

2. it's strange that it tells us about redacted *output* but not redacted *input*

```
2018/11/02 13:45:52 $ ./build/bin/fk key create
2018/11/02 13:45:52 Enter your email address, this will help other people find your key.
2018/11/02 13:45:52 [email] : 
2018/11/02 13:46:00 > paul@example.com
2018/11/02 13:46:00 Your key will be protected with this password:
2018/11/02 13:46:00 *** line redacted from log file ***
2018/11/02 13:46:00 If you use a password manager, save it there now.
2018/11/02 13:46:00 Store this safely, otherwise you won’t be able to use your key
2018/11/02 13:46:00 Press enter when you've stored it safely. 
2018/11/02 13:46:11 > 
2018/11/02 13:46:11 [H[2J
2018/11/02 13:46:11 Enter the 4th word from your password
2018/11/02 13:46:11 [4th word] : 
2018/11/02 13:46:14 Creating key for paul@example.com:
2018/11/02 13:46:14     [✔] Generate key for paul@example.com
2018/11/02 13:46:15     [✔] Store key in gpg
2018/11/02 13:46:15     [✔] Store password in system keyring
2018/11/02 13:46:15     [✔] Setup automatic maintenance using cron
2018/11/02 13:46:16     [✔] Make a backup ZIP file in
2018/11/02 13:46:16         /home/paul/.config/fluidkeys/backups/2018-11-02/
2018/11/02 13:46:16  ▸   Successfully created key for paul@example.com
```

Suggestion: if `logToFile` is `false` and the input is non-empty, instead output `*** line redacted from log file ***`

idrysdale:
Good suggestions! Will go around again on this....

PR #210, opened 2018-11-26
Introduce `fk team show <uuid>`

idrysdale:
This hits the teamserver and returns a list of members and requests to join.

In our design document, we specified that the command would be `team show` (without the `uuid` argument), as far as I can tell we've a few options as to how we'd like to solve that but I've paused to talk it through with you first @paulfurley. We might want to record which teams people are in within their `db.json` file - or perhaps another trip to the team server.

I figured in the meantime, I'd get it in working with the `uuid` arg. (The same will apply for the `approve` subcommand).

TODO: Rebase after https://github.com/fluidkeys/fluidkeys/pull/209, https://github.com/fluidkeys/fluidkeys/pull/208

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ MAIN_GO_FILES=fluidkeys/main.go \
 	     fluidkeys/privatekeys.go \
 	     fluidkeys/ui.go \
 	     fluidkeys/keyfromgpg.go \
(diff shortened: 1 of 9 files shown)


paulfurley:
Closing these for now.

PR #212, opened 2018-12-05
Implement happy path to `send secret`

idrysdale:


api/api.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+// A Client manages communication with the Fluidkeys Server API.
+type Client struct {
+	client    *http.Client // HTTP client used to communicate with the API.
+	BaseURL   *url.URL     // Base URL for API requests
+	UserAgent string       // User agent used when communicating with the  API.
+}
+
+// NewClient returns a new Fluidkeys Server API client. If a nil httpClient is
```
paulfurley: Docs mention non-existent `httpClient`

api/api.go
```diff
@@ -0,0 +1,127 @@
...
+	decodedJSON := new(v1structs.GetPublicKeyResponse)
+	response, err := c.do(request, &decodedJSON)
+	if err != nil {
+		return "", response, err
+	}
+	return decodedJSON.ArmoredPublicKey, response, nil
+}
+
+// CreateSecret creates a secret for the given recipient
+func (c *Client) CreateSecret(recipientFingerprint string, armoredEncryptedSecret string) (*http.Response, error) {
```
paulfurley: `recipientFingerprint` should be a `fingerprint.Fingerprint` type

api/api.go
```diff
@@ -0,0 +1,127 @@
...
+	sendSecretRequest := v1structs.SendSecretRequest{
+		RecipientFingerprint:   recipientFingerprint,
+		ArmoredEncryptedSecret: armoredEncryptedSecret,
+	}
+	url := fmt.Sprintf("secrets")
```
paulfurley: is this supposed to be simply `url := "secrets"` ?

api/api.go
```diff
@@ -0,0 +1,127 @@
...
+	return &Client{
+		client:    http.DefaultClient,
+		BaseURL:   baseURL,
+		UserAgent: userAgent,
+	}
+}
+
+// GetPublicKey attempts to get a single armorded public key.
+func (c *Client) GetPublicKey(email string) (string, *http.Response, error) {
+	url := fmt.Sprintf("email/%s/key", url.QueryEscape(email))
```
paulfurley: Being picky, it's not a `url` (`https://example.com/v1/email`) it's *part* of the path of a full URL. Suggest using `path` instead of `url`

api/api.go
```diff
@@ -0,0 +1,127 @@
...
+
+	response, err := c.do(request, nil)
+	if err != nil {
+		return response, err
+	}
+
+	return response, nil
+}
+
+// newRequest creates an API request. A relative URL can be provided in urlStr,
```
paulfurley: If i read the code correctly, it *only* accepts a relative path, right? If so, it could just be called `relativePath` and the docstring could say "relativePath is resolved relative to the BaseURL of the client"

paulfurley: Actually I *didn't* read the code right.

The docstring tells me about using a relative path, hinting that I could instead use an absolute path (or URL?) but I'm not sure how.

Suggest: make it require a relative path, don't allow absolute path-or-url and make it simpler.

api/api.go
```diff
@@ -0,0 +1,127 @@
...
+	if !strings.HasSuffix(c.BaseURL.Path, "/") {
+		return nil, fmt.Errorf("BaseURL must have a trailing slash, but %q does not", c.BaseURL)
+	}
+	url, err := c.BaseURL.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	var buf io.ReadWriter
+	if body != nil {
```
paulfurley: Suggest renaming `body` to `requestData` since `body` says to me "bytes to send with the request"

api/api.go
```diff
@@ -0,0 +1,127 @@
...
+		buf = new(bytes.Buffer)
+		enc := json.NewEncoder(buf)
+		enc.SetEscapeHTML(false)
```
paulfurley: ooh interesting, why do this?

api/api.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	if c.UserAgent != "" {
+		request.Header.Set("User-Agent", c.UserAgent)
+	}
+	return request, nil
+}
+
+// do sends an API request and decodes the JSON response, storing it in the
+// value pointed to by v. If an API error occurs, it returns error.
+func (c *Client) do(req *http.Request, v interface{}) (*http.Response, error) {
```
paulfurley: suggest renaming `v` to `responseData`

fluidkeys/secretsend.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	encryptedSecret, err := encryptSecret(secret, pgpKey)
+	if err != nil {
+		printFailed("Couldn't encrypt the secret:")
+		out.Print("Error: " + err.Error() + "\n")
+		return 1
+	}
+
+	response, err := client.CreateSecret(fmt.Sprintf("OPENPGP4FPR:%s", pgpKey.Fingerprint().Hex()), encryptedSecret)
```
paulfurley: Here's a case where I'd say the API boundary isn't right: it's unreasonable for `main` to have to know how to encode a fingerprint for the API. `client.CreateSecret` should take the `fingerprint.Fingerprint` type directly and do the formatting itself

fluidkeys/secretsend.go
```diff
@@ -0,0 +1,108 @@
...
+	if response.StatusCode != 201 || err != nil {
+		printFailed("Couldn't send the secret to " + recipientEmail)
+		if err != nil {
```
paulfurley: Worth noting: if the API returns a 400, 500, whatever, I've tried hard to ensure it always includes a JSON-encoded human friendly message:
```json
{
    "detail": "the database didn't respond"
}
```

... you could make a helper which tries to decode that (see `v1structs.ErrorResponse`) and possible display it

fluidkeys/secretsend.go
```diff
@@ -0,0 +1,108 @@
...
+	return 0
+}
+
+func scanUntilEOF() (message string, err error) {
+	reader := bufio.NewReader(os.Stdin)
+	var output []rune
+
+	for {
+		input, _, err := reader.ReadRune()
+		if err != nil && err == io.EOF {
```
paulfurley: this func can return an error, but as far as I can tell, there's no codepath which actually does that.

are you sure the `err != nil` and `err == io.EOF` should be handled identically?

idrysdale: This is if there's an error *and* it's an EOF error. I could add a `panic` on an error that's not EOF?

paulfurley: I think the caller already `panics` if you return an error, so returning an error seems more reasonable than panicking in here :)

paulfurley: BTW you can safely test for `err == io.EOF` without `err != nil && `

api/api.go
```diff
@@ -51,9 +53,9 @@ func (c *Client) GetPublicKey(email string) (string, *http.Response, error) {
 }
 
 // CreateSecret creates a secret for the given recipient
-func (c *Client) CreateSecret(recipientFingerprint string, armoredEncryptedSecret string) (*http.Response, error) {
+func (c *Client) CreateSecret(recipientFingerprint fingerprint.Fingerprint, armoredEncryptedSecret string) (*http.Response, error) {
 	sendSecretRequest := v1structs.SendSecretRequest{
-		RecipientFingerprint:   recipientFingerprint,
+		RecipientFingerprint:   fmt.Sprintf("OPENPGP4FPR:%s", recipientFingerprint.Hex()),
```
paulfurley: :+1:

fluidkeys/secretsend.go
```diff
@@ -107,3 +112,14 @@ func encryptSecret(secret string, pgpKey *pgpkey.PgpKey) (string, error) {
 	message.Close()
 	return buffer.String(), nil
 }
+
+func decodeErrorResponse(response *http.Response) string {
+
+	errorResponse := v1structs.ErrorResponse{}
+	if err := json.NewDecoder(os.Stdin).Decode(&errorResponse); err != nil {
```
paulfurley: `os.Stdin` ? :sneezing_face:

fluidkeys/secretsend.go
```diff
@@ -76,9 +76,12 @@ func scanUntilEOF() (message string, err error) {
 
 	for {
 		input, _, err := reader.ReadRune()
-		if err != nil && err == io.EOF {
+		if err == io.EOF {
 			break
 		}
+		if err != io.EOF {
```
paulfurley: and if `ReadRune()` returns with *no* error (`nil`) ? :bomb:

paulfurley:
I'm rebasing this

fluidkeys/secretsend.go
```diff
@@ -0,0 +1,128 @@
...
+		printFailed("Couldn't send the secret to " + recipientEmail)
+		if err != nil {
+			apiErrorResponseDetail := decodeErrorResponse(response)
+			out.Print("Error: " + err.Error() + "\n")
+			out.Print("Error: " + apiErrorResponseDetail + "\n")
+		}
+		return 1
+	}
+
+	printSuccess("Successfuly sent secret to " + recipientEmail + "\n")
```
paulfurley: "Successfully"

PR #217, opened 2018-12-07
Introduce `secret receive`

idrysdale:
This subcommand makes an API client call `client.ListSecrets`, and
then attempts to decrypt the content of these secrets against a key.

### DIFF
--- a/fluidkeys/secretreceive.go
+++ b/fluidkeys/secretreceive.go
@@ -29,6 +29,8 @@ func secretReceive() exitCode {
 
 	out.Print(colour.Info("Downloading secrets...") + "\n\n")
 
+	var sawError bool = false
+
 	for _, key := range keys {
 		secrets, secretErrors, err := downloadAndDecryptSecrets(key)
 		if err != nil {
@@ -59,10 +61,14 @@ func secretReceive() exitCode {
 			for _, error := range secretErrors {
 				printFailed(error.Error())
 			}
-			// TODO: Record the fact we saw an error, don't return 0
+			sawError = true
 		}
 	}
-	return 0
+	if sawError {
+		return 1
+	} else {
+		return 0
+	}
 }
 
 func downloadAndDecryptSecrets(key pgpkey.PgpKey) (decryptedSecrets []string, secretErrors []error, err error) {
--- a/fluidkeys/secretreceive.go
+++ b/fluidkeys/secretreceive.go
@@ -36,50 +36,52 @@ func secretReceive() exitCode {
 			printFailed("Couldn't get email for key " + key.Fingerprint().String() + "\n")
 			continue
 		}
-
-		if encryptedSecrets, err := client.ListSecrets(key.Fingerprint()); err != nil {
+		encryptedSecrets, err := client.ListSecrets(key.Fingerprint())
+		if err != nil {
 			out.Print("📪 " + displayName(&key) + ": " + colour.Failure(err.Error()) + "\n")
+			continue
+		}
+
+		if len(encryptedSecrets) == 0 {
+			out.Print("📭 " + displayName(&key) + ": No secrets found\n")
+			continue
+		}
+		var passwordPrompter promptForPasswordInterface
+		passwordPrompter = &interactivePasswordPrompter{}
+		if privateKey, _, err := getDecryptedPrivateKeyAndPassword(&key, passwordPrompter); err != nil {
+			err := fmt.Sprintf("Error getting private key and password: %s", err)
+			out.Print("📪 " + displayName(&key) + ": " + colour.Failure(err) + "\n")
+			continue
 		} else {
-			if encryptedSecrets == nil {
-				out.Print("📭 " + displayName(&key) + ": No secrets found\n")
-				continue
-			}
-			a := LoadPrivateKeyFromGnupg{}
-			if privateKey, _, err := getDecryptedPrivateKeyAndPassword(&key, a.passwordGetter); err != nil {
-				err := fmt.Sprintf("Error getting private key and password: %s", err)
-				out.Print("📪 " + displayName(&key) + ": " + colour.Failure(err) + "\n")
-				continue
+			key = *privateKey
+		}
+
+		out.Print("📬 " + displayName(&key) + ":\n")
+
+		errors := []error{}
+		for i, encryptedSecret := range encryptedSecrets {
+			decryptedContent, err := decrypt(encryptedSecret.EncryptedContent, &key)
+			if err != nil {
+				errors = append(errors, err)
 			} else {
-				key = *privateKey
-			}
+				displayCounter := strconv.Itoa(i+1) + ". "
+				barLength := secretDividerLength - len(displayCounter)
+				out.Print(displayCounter + strings.Repeat("─", barLength) + "\n")
 
-			out.Print("📬 " + displayName(&key) + ":\n")
-
-			errors := []error{}
-			for i, encryptedSecret := range encryptedSecrets {
-				decryptedContent, err := decrypt(encryptedSecret.EncryptedContent, &key)
-				if err != nil {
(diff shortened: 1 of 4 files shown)


fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			printFailed("Couldn't get email for key " + key.Fingerprint().String() + "\n")
+			continue
+		}
+
+		out.Print("For " + email + ":\n\n")
+		if encryptedSecrets, err := client.ListSecrets(key.Fingerprint()); err != nil {
+			printFailed(err.Error() + "\n")
+		} else {
+			if encryptedSecrets == nil {
```
paulfurley: Not sure why you're comparing to `nil` when you want to check the length?

What happens if `encryptedSecrets` *is* empty? Its value isn't `nil`...

See https://play.golang.org/p/3-ga7iFbLUS

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,90 @@
...
+	for _, key := range keys {
+		email, err := key.Email()
+		if err != nil {
+			printFailed("Couldn't get email for key " + key.Fingerprint().String() + "\n")
+			continue
+		}
+
+		out.Print("For " + email + ":\n\n")
+		if encryptedSecrets, err := client.ListSecrets(key.Fingerprint()); err != nil {
+			printFailed(err.Error() + "\n")
```
paulfurley: i think you'd benefit from `continue` here rather than the subsequent `else` at a new level of indentation

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,90 @@
...
+		} else {
+			if encryptedSecrets == nil {
+				printInfo("No secrets found\n")
+				continue
+			}
+			a := LoadPrivateKeyFromGnupg{}
```
paulfurley: hmm. novel! but this action wasn't intended to be used this way, sniff sniff

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,90 @@
...
+			if privateKey, _, err := getDecryptedPrivateKeyAndPassword(&key, a.passwordGetter); err != nil {
```
paulfurley: I'm not sure how `a.passwordGetter` could ever be set here? Also `passwordGetter`'s type is `promptForPasswordInterface`. I think this might explain the segfault I experienced :)

fluidkeys/secretreceive.go
paulfurley: Ah. Or bringing out `var encryptedSecrets []v1structs.EncryptedSecret` on the line above

paulfurley: But better what you said.

idrysdale: 👍

fluidkeys/secretreceive.go
idrysdale: 😅 Sorted!

fluidkeys/secretreceive.go
idrysdale: Doh!

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,122 @@
...
+		if err != nil {
+			out.Print("📪 " + displayName(&key) + ": " + colour.Failure(err.Error()) + "\n")
+			continue
+		}
+
+		if len(encryptedSecrets) == 0 {
+			out.Print("📭 " + displayName(&key) + ": No secrets found\n")
+			continue
+		}
+		var passwordPrompter promptForPasswordInterface
```
paulfurley: I think you can drop this line. You don't need to reference the interface at all, only sure that the thing you pass *satisfies* that interface, which the `interactivePasswordPrompter{}` does

idrysdale: Ah yes!

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,122 @@
...
+
+		out.Print("📬 " + displayName(&key) + ":\n")
+
+		errors := []error{}
+		for i, encryptedSecret := range encryptedSecrets {
+			decryptedContent, err := decrypt(encryptedSecret.EncryptedContent, &key)
+			if err != nil {
+				errors = append(errors, err)
+			} else {
+				displayCounter := strconv.Itoa(i+1) + ". "
```
paulfurley: Fancy! Is there a reason not to use the (less fancy) `fmt.Sprintf("%d. ", i+1)` ?

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,122 @@
...
+			continue
+		}
+		var passwordPrompter promptForPasswordInterface
+		passwordPrompter = &interactivePasswordPrompter{}
+		if privateKey, _, err := getDecryptedPrivateKeyAndPassword(&key, passwordPrompter); err != nil {
+			err := fmt.Sprintf("Error getting private key and password: %s", err)
+			out.Print("📪 " + displayName(&key) + ": " + colour.Failure(err) + "\n")
+			continue
+		} else {
+			key = *privateKey
```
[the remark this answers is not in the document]
idrysdale: You're right, it's misleading, I'll follow the pattern above, pulling out `err` check onto new line so as to keep `privateKey` in scope.

fluidkeys/secretreceive.go
idrysdale: Nope! That's clearer.

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,122 @@
...
+			decryptedContent, err := decrypt(encryptedSecret.EncryptedContent, &key)
+			if err != nil {
+				errors = append(errors, err)
+			} else {
+				displayCounter := strconv.Itoa(i+1) + ". "
+				barLength := secretDividerLength - len(displayCounter)
+				out.Print(displayCounter + strings.Repeat("─", barLength) + "\n")
+
+				scanner := bufio.NewScanner(strings.NewReader(decryptedContent))
+				for scanner.Scan() {
```
paulfurley: I'm not sure at all what's going on here!

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,122 @@
...
+	keys, err := loadPgpKeys()
+
+	if err != nil {
+		printFailed("Couldn't load PGP keys")
+		return 1
+	}
+
+	out.Print(colour.Info("Downloading secrets...") + "\n\n")
+
+	for _, key := range keys {
```
paulfurley: the level of nesting is a bit overwhelming and i suspect this would be the easiest place to factor something out, along the lines of:
```
secrets, errors, err := receiveSecretsForKey(key)
```

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,122 @@
...
+			key = *privateKey
+		}
+
+		out.Print("📬 " + displayName(&key) + ":\n")
+
+		errors := []error{}
+		for i, encryptedSecret := range encryptedSecrets {
+			decryptedContent, err := decrypt(encryptedSecret.EncryptedContent, &key)
+			if err != nil {
+				errors = append(errors, err)
```
paulfurley: another good candidate for `continue` i think

fluidkeys/secretreceive.go
```diff
@@ -0,0 +1,122 @@
...
+			} else {
```
paulfurley: if this whole clause is just printing a repeating bar I think it'd factor out into a function beautifully :)

fluidkeys/secretreceive.go
paulfurley: I've read the docs and I'm still not sure. My best guess: is it prepending a space before every line?

PR #218, opened 2018-12-08
Delete secrets when downloading

idrysdale:
I've decided to delete a secret on the server whether it successfully
decrypts or not. I'm not sure what good it'd be to keep it on the server if
it's failed to decrypt?

Also, if we fail to either decode the metadata (for the UUID), or to
actually delete the secret on the server I've opted to log this.

### DIFF
--- a/fluidkeys/secretreceive.go
+++ b/fluidkeys/secretreceive.go
@@ -2,11 +2,14 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"strconv"
 	"strings"
 
+	"github.com/fluidkeys/api/v1structs"
 	"github.com/fluidkeys/fluidkeys/humanize"
 
 	"github.com/fluidkeys/fluidkeys/colour"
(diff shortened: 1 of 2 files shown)


fluidkeys/secretreceive.go
```diff
@@ -85,6 +88,16 @@ func downloadAndDecryptSecrets(key pgpkey.PgpKey) (decryptedSecrets []string, se
 		} else {
 			decryptedSecrets = append(decryptedSecrets, decryptedContent)
 		}
+		jsonMetadata, err := decrypt(encryptedSecret.EncryptedMetadata, privateKey)
```
paulfurley: missing an `err` check here

idrysdale: Good spot. Fixed!

paulfurley:
I've just been playing with this and the fact that it doesn't prompt caught me out: my first key got secrets OK, but then it prompted for a password on the second one. While I was messing around looking for the password, I think it deleted the secrets for the first one, since they weren't there when I ran it again!

It's a little disorientating having it (silently) delete secrets while at the same time prompting for passworsd

paulfurley:
I'm gonna merge to crack on, but add a card that we could do this with a prompt. Nice one! :)

PR #225, opened 2018-12-11
Licence with GNU Affero General Public License 3.0

paulfurley:
* add copyright / licence notice to *.go
* humanize, colour: move LICENSE into .go files
  This is so we can prevent our licence-adding script from adding to them.

* add script/licence to generate copyright headers

https://www.gnu.org/licenses/gpl-howto.html

> To do so, attach the following notices to the program. It is safest to
> attach them to the start of each source file to most effectively state
> the exclusion of warranty; and each file should have at least the
> "copyright" line and a pointer to where the full notice is found.


idrysdale:
🍸

PR #243, opened 2018-12-13
secret receive: show if a key isn't published

paulfurley:


fluidkeys/secretreceive.go
```diff
@@ -51,6 +51,11 @@ func secretReceive() exitCode {
 	var sawError bool = false
 
 	for _, key := range keys {
+		if !Config.ShouldPublishToAPI(key.Fingerprint()) {
+			message := "Not published in Fluidkeys directory"
+			out.Print("📪 " + displayName(&key) + ": " + colour.Warning(message) + "\n")
```
idrysdale: Suggest that we don't use the mailbox emoji... perhaps rather ⛔ ?

paulfurley: nice yeah

PR #256, opened 2018-12-19
clean up main package

idrysdale:
Namely:
* Remove the redundant case of 'init' command
* Return exitCode from secretSubcommand to caller (rather than exiting from there)
* Remove a stray line of commented code
* Correctly indent code (don't use spaces)


paulfurley:
Excellent, good spots!

PR #257, opened 2018-12-19
Factor RoughlyValidateEmail out and use on key create

idrysdale:


paulfurley:
Nice. Will start using myself!

PR #259, opened 2018-12-20
Introduce verify email step during key create

idrysdale:
We no longer prompt the user whether to publish their key now, too, as they won't be able to receive secrets if it's not published.

When polling the API to verify that the email matches the key in the public API, we call it every 5 seconds, swallow *any* errors, and give up after 15 minutes.


fluidkeys/ui.go
```diff
@@ -45,3 +47,7 @@ func printSuccessfulAction(message string) {
 func printFailedAction(message string) {
 	out.Print("    [" + colour.Failure("✘") + "] " + message + "\n")
 }
+
+func printHeader(message string) {
+	out.Print(colour.Header(fmt.Sprintf(" %-79s", message)) + "\n\n")
```
paulfurley: oooh fancy format string :)

paulfurley:
:heart: :heart: :heart:

PR #261, opened 2018-12-20
Prompt to delete secrets

idrysdale:
If we fail to delete a secret, we log it, but carry on.

### DIFF
--- a/fluidkeys/secretreceive.go
+++ b/fluidkeys/secretreceive.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 
 	"github.com/fluidkeys/api/v1structs"
+	"github.com/fluidkeys/fluidkeys/fingerprint"
 	"github.com/fluidkeys/fluidkeys/humanize"
 
 	"github.com/fluidkeys/fluidkeys/colour"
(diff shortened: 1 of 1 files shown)


fluidkeys/secretreceive.go
```diff
@@ -87,25 +89,40 @@ func secretReceive() exitCode {
...
 	if sawError {
 		return 1
-	} else {
-		return 0
 	}
+	return 0
 }
 
-func downloadAndDecryptSecrets(key pgpkey.PgpKey) (decryptedSecrets []string, secretErrors []error, err error) {
+func downloadAndDecryptSecrets(key pgpkey.PgpKey) (decryptedSecrets []string, decryptedUUIDs []string, secretErrors []error, err error) {
```
paulfurley: Nice for getting this working.

What do you think about this simpler alternative? return `secrets []secret, ...` where `secret` is:
```
type secret struct {
    decryptedContent string
    UUID uuid.UUID
    sentToFingerprint *fingerprint.Fingerprint
}
```

fluidkeys/secretreceive.go
```diff
@@ -190,3 +227,5 @@ type errDecryptPrivateKey struct {
 }
 
 func (e errDecryptPrivateKey) Error() string { return e.originalError.Error() }
+
+type downloadedSecrets map[fingerprint.Fingerprint][]string
```
paulfurley: is this still meant to be here?

paulfurley:
Cool! :)

fluidkeys/secretreceive.go
idrysdale: Good spot, no!

PR #262, opened 2018-12-20
Introduce `fk setup`

idrysdale:
This wraps `key create` with a friendly welcome, and then sends them a test secret.

### DIFF
--- a/fluidkeys/main.go
+++ b/fluidkeys/main.go
@@ -132,7 +132,8 @@ func keySubcommand(args docopt.Opts) exitCode {
 		"create", "from-gpg", "list", "maintain", "publish",
 	}) {
 	case "create":
-		os.Exit(keyCreate())
+		exitCode, _ := keyCreate()
+		os.Exit(exitCode)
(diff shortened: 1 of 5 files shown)


fluidkeys/setup.go
```diff
@@ -0,0 +1,66 @@
+package main
+
+import (
+	"time"
+
+	"github.com/fluidkeys/fluidkeys/colour"
+	"github.com/fluidkeys/fluidkeys/out"
+)
+
+func setup() exitCode {
```
paulfurley: this is *tidy*

fluidkeys/main.go
```diff
@@ -63,6 +63,7 @@ Configuration file: %s
 Usage:
 	fk secret send <recipient-email-address>
 	fk secret receive
+	fk setup
```
paulfurley: seems like this should be at the top, shouldn't it?

idrysdale: Yep!

PR #265, opened 2018-12-21
Print a friendly invite message if the key is not found in api

idrysdale:
Merge https://github.com/fluidkeys/www.fluidkeys.com/pull/66 when we release `fk setup` in 0.3

### DIFF
--- a/fluidkeys/secretsend.go
+++ b/fluidkeys/secretsend.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/fluidkeys/crypto/openpgp"
 	"github.com/fluidkeys/crypto/openpgp/armor"
+	"github.com/fluidkeys/fluidkeys/api"
 	"github.com/fluidkeys/fluidkeys/colour"
 	"github.com/fluidkeys/fluidkeys/out"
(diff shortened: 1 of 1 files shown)


fluidkeys/secretsend.go
```diff
@@ -27,14 +27,31 @@ import (
...
+		if err == api.ErrPublicKeyNotFound {
+			out.Print("\n")
+			out.Print("Couldn't find " + recipientEmail + " in the Fluidkeys Directory.\n\n")
+			out.Print("You can invite them to install Fluidkeys:\n")
+			out.Print("───\n")
+			out.Print(colour.Warning(`I'd like to send you an encrypted secret with Fluidkeys.
+
+You can download and set up Fluidkeys here:
+
+https://getstarted.fluidkeys.com#` + recipientEmail + `
```
paulfurley: We don't currently have `getstarted.fluidkeys.com`, nor do we have any JS that actually manipulates the instructions. 

ATM `https://download.fluidkeys.com#test`redirects to `https://www.fluidkeys.com/download/#test` so we could implement that jsmagic there?

So replace `getstarted.` with `download.` ?

PR #267, opened 2018-12-21
Begin writing out 'publish', instead talking about 'registering'

idrysdale:
Makes me think, is it time to change `fk key publish` to `fk key register`?


paulfurley:
This is awesome!

The only issue is that it's misleading if they have multiple email addresses on a key...

PR #279, opened 2019-01-11
Improve GnuPG log output

paulfurley:
* GnuPG.run(..): if gpg returns an error, return a more descriptive error message, and log
  all stderr lines to debug.log

* combine .run(..) and runWithStdin(..) into single method

### DIFF
--- a/gpgwrapper/gpgwrapper.go
+++ b/gpgwrapper/gpgwrapper.go
@@ -79,7 +79,7 @@ func Load() (*GnuPG, error) {
 
 // Returns the GnuPG version string, e.g. "1.2.3"
 func (g *GnuPG) Version() (string, error) {
-	outString, err := g.run("--version")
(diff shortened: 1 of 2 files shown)


gpgwrapper/gpgwrapper.go
```diff
@@ -336,7 +336,32 @@ func (g *GnuPG) runWithStdin(textToSend string, arguments ...string) (
...
+			extraErr = fmt.Sprintf(", stderr: %s [see fluidkeys log for more]", stderrLines[0])
+		}
+
+		log.Printf("command failed: `gpg %s` : %s", strings.Join(fullArguments, " "), err)
+		for _, line := range stderrLines {
+			log.Print(line)
+		}
+
+		returnErr = fmt.Errorf("%v%s", err, extraErr)
 		return
```
paulfurley: we decided not to use naked returns

PR #287, opened 2019-01-28
stop printing squirrel's secret in organge

idrysdale:
It was confusing a user as they then expected other secrets to also
appear in orange.


paulfurley:
Tested locally, looks great.

PR #288, opened 2019-01-29
factor out top-level fk code into `fk` package

paulfurley:
* move all of `fk` code out of `fluidkeys/*.go` (`master` package) and
into `./fk/*.go`
* make a new `cmd/fk/main.go` which *only* passes through to `fk.Main()`
* simplify `Makefile`: build now requires single file `cmd/fk/main.go`


paulfurley:
@rubdos This one's partly for you! It shouldn't affect the AUR package singe `make` is unchanged, but it's another simplification of the `Makefile` :)

PR #289, opened 2019-01-29
Send files with `fk secret send`

idrysdale:


fluidkeys/secretsend.go
```diff
@@ -104,6 +102,56 @@ https://download.fluidkeys.com#` + recipientEmail + `
...
+	return nil, fmt.Errorf("couldn't find file " + filename)
+}
+
+func fileExists(filename string) bool {
+	if _, err := os.Stat(filename); err == nil {
+		return true
+	} else if os.IsNotExist(err) {
+		return false
+	} else {
+		log.Panic()
```
paulfurley: What does this case mean? How does it happen?

fluidkeys/secretsend.go
```diff
@@ -64,29 +66,25 @@ https://download.fluidkeys.com#` + recipientEmail + `
...
-
-	if strings.TrimSpace(secret) == "" {
-		printFailed("Exiting due to empty message.\n")
+	if err != nil {
+		printFailed("Error: " + err.Error())
 		return 1
 	}
 
-	encryptedSecret, err := encryptSecret(secret, pgpKey)
+	encryptedSecret, err := encryptSecret(*secret, pgpKey, filename)
```
paulfurley: you're gonna want to convert this into *basename* somewhere, e.g. if it's `/tmp/secret.txt`, only encode `secret.txt`

fluidkeys/secretsend.go
```diff
@@ -121,18 +169,22 @@ func scanUntilEOF() (message string, err error) {
 	return string(output), nil
 }
 
-func encryptSecret(secret string, pgpKey *pgpkey.PgpKey) (string, error) {
+func encryptSecret(secret string, pgpKey *pgpkey.PgpKey, filename string) (string, error) {
```
paulfurley: Any reason you chose to put `filename` after `pgpKey` rather than before?

paulfurley:
Had to bail in the middle of reviewing this, will look tomorrow morning!

paulfurley:
```

File: fk/main.go [hunk 4]
```diff
@@ -65,7 +65,7 @@ Configuration file: %s
 Usage:
 	fk setup
 	fk setup <email>
-	fk secret send <recipient-email-address>
+	fk secret send <recipient-email-address> [--file <filename>]
```
paulfurley: I think you meant `[--file=<filename>]`

From the [documentation](http://docopt.org/):

```
Usage:
  naval_fate ship new <name>...
  naval_fate ship <name> move <x> <y> [--speed=<kn>]
  naval_fate ship shoot <x> <y>
  naval_fate mine (set|remove) <x> <y> [--moored|--drifting]
  naval_fate -h | --help
  naval_fate --version
```

We want the equivalent to `[--speed=<kn>]`

More generally though, I found it unnatural specifying the filename *after* the recipient.   I wanted to do:

```
fk secret send /tmp/secret.txt ...
```

Which makes me wonder if it should be:
```
fk secret send <recipient-email-address>                                # stdin version
fk secret send <filename> --to=<recipient-email-address>
```
?

fk/secretsend.go
```diff
@@ -121,18 +169,22 @@ func scanUntilEOF() (message string, err error) {
...
 
-func encryptSecret(secret string, pgpKey *pgpkey.PgpKey) (string, error) {
+func encryptSecret(secret string, pgpKey *pgpkey.PgpKey, filename string) (string, error) {
 	buffer := bytes.NewBuffer(nil)
 	message, err := armor.Encode(buffer, "PGP MESSAGE", nil)
 	if err != nil {
 		return "", err
 	}
 
+	fileHints := openpgp.FileHints{
```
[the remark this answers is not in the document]
paulfurley: where `true` is actually `false`.

fluidkeys/secretsend.go
idrysdale: I got into a bit of a rabbit hole with this, having read https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go

I now think that a better approach is to completely abandon testing if this file exists at this point in the code, as it's racy: the file may exist when we test, but may not when we go to open it. Instead I think we should simply try to open the file and if it fails, output the error we get.

fluidkeys/secretsend.go
idrysdale: You're right!

idrysdale:
Ok, thanks for the time reviewing that @paulfurley. I've addressed your points with subsequent commits, plus added a bit to the tests.

This is ready for another review! Thanks! 🙌

fk/secretsend.go
```diff
@@ -149,4 +187,19 @@ func encryptSecret(secret string, pgpKey *pgpkey.PgpKey) (string, error) {
 	return buffer.String(), nil
 }
 
+func makeFileHintsForFilename(filename string) *openpgp.FileHints {
```
paulfurley: I was going to suggest returning a `FileHints` instead a `*FileHints`.

But I see that would make your caller-code slightly less compact, and given that `nil` is an acceptable argument into `openpgp.Encrypt(...)` I take it back !

fk/secretsend_test.go
```diff
@@ -21,10 +21,34 @@ func TestEncryptSecret(t *testing.T) {
...
+
+		messageDetails := decryptMessageDetails(armoredEncryptedSecret, pgpKey, t)
+		assertMessageBodyMatchesSecretContent(messageDetails.UnverifiedBody, secret, t)
+		assert.Equal(t, "_CONSOLE", messageDetails.LiteralData.FileName)
+		if messageDetails.LiteralData.ForEyesOnly() != true {
+			t.Fatalf("expected secret to be For Eyes Only, but isn't")
+		}
+	})
+
+	t.Run("with a filename", func(t *testing.T) {
```
paulfurley: what happens for an empty *file* ?

fk/secretsend.go
```diff
@@ -64,29 +67,28 @@ https://download.fluidkeys.com#` + recipientEmail + `
...
-	out.Print(colour.Info("   It will be end-to-end encrypted so no-one else can read it\n\n"))
-
-	secret, err := scanUntilEOF()
-	if err != nil {
-		log.Panic(err)
-		return 1
+	var secret *string
+	var basename string
+	if filename != "" {
+		secret, err = getSecretFromFile(filename)
```
paulfurley: Is this a tab / space issue? I'll look into it!

paulfurley: It seems this is *fixing* the code by making this a tab. One of us needs to get our editor to use `gofmt` ... and I think it's me...

fk/secretsend_test.go
paulfurley: (or one with only whitespace)

paulfurley: image

paulfurley: I'm gonna add this as a new bug so we can get this merged.

paulfurley:
It crashes if you answer "no" :(

```
$ ffk secret send --to paul@fluidkeys.com /tmp/secret 
go build -o build/bin/fk cmd/fk/main.go
 ▸   Found public key for paul@fluidkeys.com
---
---

Send /tmp/secret? [Y/n] n

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x76e3ef]

goroutine 1 [running]:
github.com/fluidkeys/fluidkeys/fk.secretSend(0x7ffcbf478f9e, 0x12, 0x7ffcbf478fb1, 0xb, 0xb)
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/fk/secretsend.go:91 +0x4cf
github.com/fluidkeys/fluidkeys/fk.secretSubcommand(0xc42015a690, 0xc420394b70)
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/fk/main.go:322 +0x29c
github.com/fluidkeys/fluidkeys/fk.Main(0xc420030118)
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/fk/main.go:96 +0x35d
main.main()
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/cmd/fk/main.go:26 +0x22

```

fk/secretsend.go
```diff
@@ -104,6 +106,42 @@ https://download.fluidkeys.com#` + recipientEmail + `
...
+		return nil, fmt.Errorf("error reading file: " + err.Error())
+	}
+	secret := string(secretData)
+	out.Print("---\n")
+	out.Print(secret)
+	out.Print("---\n\n")
+
+	prompter := interactiveYesNoPrompter{}
+
+	if prompter.promptYesNo("Send "+filename+"?", "y", nil) {
```
paulfurley: This is the cause of the crash. If you answer "no" it just returns a `nil` string pointer and *nil error*.

I didn't spot this from view code because *normally* we use the final return as a *successful* return, and previous returns as failure cases.

I'd strongly suggest flipping the logic here to test is `promptYesNo` is `false`, and use the final return as the success argument.

(and as per my other comment, return a `string` not a `*string`)

fluidkeys/secretsend.go
paulfurley: Cool.

In that case, definitely put something inside the `Panic()` so the user doesn't just have a mystery crash with no message :)

fk/secretsend.go
idrysdale: Thanks!

fk/secretsend.go
idrysdale: :-/

fk/secretsend_test.go
idrysdale: Thanks! I've picked this up, but can drop commit if you'd rather!

fk/secretsend.go
```diff
@@ -104,6 +106,42 @@ https://download.fluidkeys.com#` + recipientEmail + `
 	return 0
 }
 
+func getSecretFromFile(filename string) (*string, error) {
```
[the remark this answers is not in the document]
idrysdale: You're right!

PR #291, opened 2019-02-01
Don't send secrets that contain a disallowed rune

idrysdale:
We disallow  all terminal control codes, *except* line feed and carriage return.

We allow these two, since if I want to send a secret to another person, I'd expect the
content to retain it's original format with line breaks.


fk/secretsend_test.go
```diff
@@ -79,3 +88,266 @@ func assertMessageBodyMatchesSecretContent(unverifiedBody io.Reader, secret stri
...
 }
+
+const (
+	englishAlphabet     = "AaBbCcDdEeFfGgHhIiJjKkLlMmNnOoPpQqRrSsTtUuVvWwXxYyZz"
+	printableCharacters = "☠☮☯♠Ω♤♣♧♥♡♦♢♔♕♚♛⚜★☆✮✯☄☾☽☼☀☁☂☃☻☺☹۞۩ εїзƸ̵̡Ӝ̵̨̄ƷξЖЗεжз☎☏¢☚☛"
+	sampleEmojis        = "😀😁😂🤣😃😄😅😆😉😊😋😎😍😘😗😙😚🙂"
+	breaks              = "line 1\rline 2\n"
+)
+
+func TestContainsDisallowedRunes(t *testing.T) {
```
paulfurley: Because the function signature is not a valid test function, these tests don't actually run

paulfurley: errr

paulfurley: I'm talking bollocks, ignore me.

idrysdale: 🤣

fk/secretsend_test.go
```diff
@@ -79,3 +88,266 @@ func assertMessageBodyMatchesSecretContent(unverifiedBody io.Reader, secret stri
...
+	for name, code := range validANSICodes {
+		t.Run(fmt.Sprintf("for '%v')", name), func(t *testing.T) {
+			if containsDisallowedRune(code) != false {
+				t.Fatalf("expected '%v' (%v) to be allowed but it wasn't", code, name)
+			}
+		})
+	}
+
+	for _, naughtyString := range naughtystrings.Unencoded() {
+		t.Run(fmt.Sprintf("'%v' doesn't crash it", naughtyString), func(t *testing.T) {
```
paulfurley: I know I suggested using a new test runner for each string but I regret it since it spams the rest output in a way that's not particularly helpful!

paulfurley:
Related: https://trello.com/c/IgjWRBtO/647-secrets-block-receiving-of-secrets-containing-invalid-runes

PR #293, opened 2019-02-05
Iterate over the secrets on `secret receive`

idrysdale:
Following the designs on the [Command Line Google Doc](https://docs.google.com/document/d/1FHQXHP7X8Uvtvn72OqZNVYyFMocrJ2MM1zWtYTM4d0c/#heading=h.435am9i821s9) `secret receive` now prints the number of secrets each key is receiving (and no longer list them numerically). It then iterates over each secret, prompting the user whether they want to copy the contents to the clipboard, then whether they want to delete it.

I opted to introduce https://github.com/atotto/clipboard/ to handle copying to the clipboard. It handles unix, with a friendly error message if they don't have a command
line tool installed:

screenshot 2019-02-05 at 11 27 11

### DIFF
--- a/fk/secretreceive.go
+++ b/fk/secretreceive.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/atotto/clipboard"
 	"github.com/fluidkeys/api/v1structs"
 	"github.com/fluidkeys/fluidkeys/colour"
 	"github.com/fluidkeys/fluidkeys/fingerprint"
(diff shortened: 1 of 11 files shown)


paulfurley:
I'll rebase

paulfurley:
These are very minor suggestions that could go in a separate PR:

paulfurley:
Another minor suggestion, it would be nice to be able to type "d" at the "Copy to clipboard" prompt to say "no, and just delete it"

fk/secretreceive.go
```diff
@@ -84,10 +84,15 @@ func secretReceive() exitCode {
 
 		for _, secret := range decryptedSecrets {
 			out.Print(formatSecretListItem(secret.decryptedContent))
+			if prompter.promptYesNo("Delete now?", "Y", nil) == true {
+				err := client.DeleteSecret(secret.sentToFingerprint, secret.UUID.String())
```
paulfurley: seems we no longer need `secret.sentToFingerprint` if we're deleting at this point... right?

fk/secretreceive.go
```diff
@@ -84,6 +85,12 @@ func secretReceive() exitCode {
 
 		for _, secret := range decryptedSecrets {
 			out.Print(formatSecretListItem(secret.decryptedContent))
+			if prompter.promptYesNo("Copy to clipboard?", "", nil) == true {
```
paulfurley: it's cool, this

paulfurley:
Awesome! Let's try and keep the `Gopkg.*` commits separate from code changes

This is cool!

PR #294, opened 2019-02-05
Prompt to save a secret to disk when receiving a file

idrysdale:
Following the designs on the [Command Line Google Doc](https://docs.google.com/document/d/1FHQXHP7X8Uvtvn72OqZNVYyFMocrJ2MM1zWtYTM4d0c/#heading=h.435am9i821s9) `secret receive` will now recieve secrets sent as files! 🙌

It will save files in `~\Downloads` and report an error if this directory doesn't exist.

If the file already exists, it'll attempt to increment a counter to find a filepath that doesn't exist. So if trying to download `README.md` and that file already exists, it'll suggest `README(1).md`.

Resulting interface:

screenshot 2019-02-05 at 17 43 21

Rebase after https://github.com/fluidkeys/fluidkeys/pull/293

### DIFF
--- a/fk/secretreceive.go
+++ b/fk/secretreceive.go
@@ -95,17 +95,13 @@ func secretReceive() exitCode {
 				secret.decryptedContent, secret.originalFilename),
 			)
 			if secret.originalFilename != "" {
-				filenameToWrite, err := getFileNameToWrite(secret.originalFilename)
+
+				err := promptAndWriteToDownloads(secret, &prompter)
 				if err != nil {
-					printFailed("Error prompting to save file to disk:")
+					printFailed("Error saving file:")
 					printFailed(err.Error())
 				}
-				if prompter.promptYesNo("Save to "+filenameToWrite+"?", "", nil) == true {
-					err := ioutil.WriteFile(filenameToWrite, []byte(secret.decryptedContent), 0644)
-					if err != nil {
-						printFailed(err.Error())
-					}
-				}
+
 			} else {
 				if prompter.promptYesNo("Copy to clipboard?", "", nil) == true {
 					err := clipboard.WriteAll(secret.decryptedContent)
@@ -138,6 +134,29 @@ func secretReceive() exitCode {
 		return 1
 	}
 	return 0
+}
+
+func promptAndWriteToDownloads(secret secret, prompter promptYesNoInterface) error {
+	downloadsDir, err := getDownloadsDir()
+	if err != nil {
+		return fmt.Errorf("Error getting downloads directory: %v", err)
+	}
+
+	filename, err := getAvailableFilename(
+		downloadsDir, secret.originalFilename, &fileSafeToWriteChecker{})
+
+	if err != nil {
+		return fmt.Errorf("Error finding available filename in %s: %v", downloadsDir, err)
+	}
+
+	if prompter.promptYesNo("Save to "+filename+"?", "", nil) == true {
+		err := ioutil.WriteFile(filename, []byte(secret.decryptedContent), 0644)
+
+		if err != nil {
+			return fmt.Errorf("Error writing file %s: %v", filename, err)
+		}
+	}
+	return nil
 }
 
 func downloadEncryptedSecrets(fingerprint fp.Fingerprint, secretLister listSecretsInterface) (
@@ -232,7 +251,43 @@ func countDigits(i int) (count int) {
 	return len(iString)
 }
 
-func getFileNameToWrite(filename string) (string, error) {
+// getAvailableFilename looks in the downloads directory for a non-existent file with the
+// requestedFilename. If the filename exists, it tries numbered alternatives, e.g.
+// `secret.txt` -> `~/Downloads/secret.txt` or `~/Downloads/secret(1).txt`
+func getAvailableFilename(
+	directory string,
+	requestedFilename string,
+	checker fileSafeToWriteInterface) (string, error) {
+
+	possibleBasenames := generateIncrementedFilenames(requestedFilename)
+	for _, possibleBasename := range possibleBasenames {
+		filenameToWrite := filepath.Join(directory, possibleBasename)
+
+		if checker.IsSafeToWrite(filenameToWrite) {
+			return filenameToWrite, nil
+		}
+	}
+	return "", fmt.Errorf("tried %s, %s, %s...",
+		possibleBasenames[0], possibleBasenames[1], possibleBasenames[2])
+}
+
+// generateIncrementedFilenames returns 11 possible basenames e.g.
+// ["file.txt", "file(1).txt", "file(2).txt" .. "file(10).txt"]
+func generateIncrementedFilenames(basename string) []string {
+	basenames := []string{
+		basename,
+	}
+
+	for i := 1; i <= 10; i += 1 {
+		beforeDot, afterDot := splitFileExtension(basename)
+
+		numberedFilename := fmt.Sprintf("%s(%d)%s", beforeDot, i, afterDot)
+		basenames = append(basenames, numberedFilename)
+	}
+	return basenames
+}
+
+func getDownloadsDir() (string, error) {
 	userDir, err := homedir.Dir()
 	if err != nil {
 		return "", err
@@ -247,43 +302,15 @@ func getFileNameToWrite(filename string) (string, error) {
 		}
 	}
 	if !fileInfo.IsDir() {
-		return "", fmt.Errorf(fileInfo.Name() + " file exists and is not a directory")
+		return "", fmt.Errorf(fileInfo.Name() + " file exists but is not a directory")
 	}
-	filenameToWrite := getNewUniqueFilename(
-		filepath.FromSlash(downloadsDir+"/"+filename),
-		doesntExist,
-	)
-
-	return filenameToWrite, nil
+	return downloadsDir, nil
 }
 
-func getNewUniqueFilename(fp string, checker func(string) bool) string {
-	i := 0
-	foundFilename := false
-	filename := filepath.Base(fp)
-	numberedFilename := filename
-	for ok := true; ok; ok = (foundFilename == false) {
-		if i == 0 {
-			// try using the original filename it was sent as first with no count suffix
-			numberedFilename = filename
-		} else if i > 0 {
-			// the original filename already exists, try `filename(i).ext` next...
-			numberedFilename = fmt.Sprintf(
-				"%s(%d)%s",
-				FilenameWithoutExtension(filename), i, path.Ext(filename),
-			)
-		}
-		if checker(filepath.FromSlash(filepath.Dir(fp) + "/" + numberedFilename)) {
-			foundFilename = true
-		} else {
-			i++
-		}
-	}
-	return numberedFilename
-}
+type fileSafeToWriteChecker struct{}
 
-func doesntExist(path string) bool {
-	if _, err := os.Stat(path); err != nil {
+func (f *fileSafeToWriteChecker) IsSafeToWrite(fullFilename string) bool {
+	if _, err := os.Stat(fullFilename); err != nil {
 		if os.IsNotExist(err) {
 			return true
 		}
@@ -291,8 +318,10 @@ func doesntExist(path string) bool {
 	return false
 }
 
-func FilenameWithoutExtension(fn string) string {
-	return strings.TrimSuffix(fn, path.Ext(fn))
+func splitFileExtension(basename string) (string, string) {
+	extension := path.Ext(basename)
+
+	return strings.TrimSuffix(basename, extension), extension
 }
 
 const (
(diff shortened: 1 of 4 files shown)


idrysdale:
I've rebased!

paulfurley:
Note that "well known" directories like "~/Downloads" are referred to as [XDG user directories](https://wiki.archlinux.org/index.php/XDG_user_directories

Long story short: we allow users to override the location using the `XDG_DOWNLOAD_DIR` environment variable (defaulting to `$HOME/Downloads` if not set is OK)

fk/secretreceive.go
```diff
@@ -206,13 +234,77 @@ func countDigits(i int) (count int) {
 	return len(iString)
 }
 
+func getFileNameToWrite(filename string) (string, error) {
```
paulfurley: This would be clearer if it were called e.g. `getAbsoluteFilenameInDownloadsDir(basename string)`

paulfurley:
Gah I'm gonna have to finish looking at this tomorrow!

paulfurley:
I've split out the 2 changes to `PgpKey.Decrypt*` as 2 different commits

fk/secretreceive.go
```diff
@@ -84,12 +91,26 @@ func secretReceive() exitCode {
...
-			if prompter.promptYesNo("Copy to clipboard?", "", nil) == true {
-				err := clipboard.WriteAll(secret.decryptedContent)
+			out.Print(formatSecretListItem(secret.decryptedContent, secret.filename))
+			if secret.filename != "" {
+				filenameToWrite, err := getFileNameToWrite(secret.filename)
 				if err != nil {
+					printFailed("Error prompting to save file to disk:")
 					printFailed(err.Error())
 				}
+				if prompter.promptYesNo("Save to "+filenameToWrite+"?", "", nil) == true {
```
paulfurley: I would've expected this to show the *full* download path:

```
🤫─────────────────────────────
Filename: secret.txt
73599be2-2abd-11e9-a5fc-ebb0c16d6c4b
──────────────────────────────
Save to secret.txt? [y/n] 
```

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
 	return len(iString)
 }
 
+func getFileNameToWrite(filename string) (string, error) {
+	userDir, err := homedir.Dir()
+	if err != nil {
+		return "", err
+	}
+	downloadsDir := userDir + "/Downloads"
```
paulfurley: This should be `filepath.Join(..)` (it's generally unsafe to concatenate directory separators as strings, and not cross-platform, in case we ever supported Windows)

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+	fileInfo, err := os.Stat(downloadsDir)
+	if err != nil {
```
paulfurley: this would be nicely factored out into e.g.
````
err := ensureDirectoryExists(downloadsDir)
if err != nil {
   ...
}
```

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+			return "", fmt.Errorf(downloadsDir + " directory doesn't exist")
+		} else {
+			return "", err
+		}
+	}
+	if !fileInfo.IsDir() {
+		return "", fmt.Errorf(fileInfo.Name() + " file exists and is not a directory")
+	}
+	filenameToWrite := getNewUniqueFilename(
+		filepath.FromSlash(downloadsDir+"/"+filename),
```
paulfurley: Novel! But best just stick to `filepath.Join(downloadsDir, filename)`

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+			// try using the original filename it was sent as first with no count suffix
+			numberedFilename = filename
+		} else if i > 0 {
+			// the original filename already exists, try `filename(i).ext` next...
+			numberedFilename = fmt.Sprintf(
+				"%s(%d)%s",
+				FilenameWithoutExtension(filename), i, path.Ext(filename),
+			)
+		}
+		if checker(filepath.FromSlash(filepath.Dir(fp) + "/" + numberedFilename)) {
```
paulfurley: `filepath.Join(..)`

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+			foundFilename = true
+		} else {
+			i++
+		}
+	}
+	return numberedFilename
+}
+
+func doesntExist(path string) bool {
```
paulfurley: Yep I see what you're doing: I think it would be clearer if this was positive though: `fileExists` (and I think we may have that code elsewhere, so ripe for a utility)

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+	}
+	filenameToWrite := getNewUniqueFilename(
+		filepath.FromSlash(downloadsDir+"/"+filename),
+		doesntExist,
+	)
+
+	return filenameToWrite, nil
+}
+
+func getNewUniqueFilename(fp string, checker func(string) bool) string {
```
paulfurley: `fp` leaves me a little cold: is that *full filename* ? or *base filename*? Or maybe *file pointer* ?

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+	i := 0
+	foundFilename := false
+	filename := filepath.Base(fp)
+	numberedFilename := filename
+	for ok := true; ok; ok = (foundFilename == false) {
+		if i == 0 {
```
paulfurley: Rather than having the `0` case exception *inside* the for loop you might want to just have a check before it (and before you set up `i`, `numberedFilename` e.g.:

```
if !fileExists(filenamePassedIntoFunction) {
    return filenamePassedIntoFunction
}
```

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+
+	return filenameToWrite, nil
+}
+
+func getNewUniqueFilename(fp string, checker func(string) bool) string {
+	i := 0
+	foundFilename := false
+	filename := filepath.Base(fp)
+	numberedFilename := filename
+	for ok := true; ok; ok = (foundFilename == false) {
```
paulfurley: I read a rule once from a reliability engineer that read something like "never let anything be unbounded / always limit the number of attempts". I think it applies here: if something strange went wrong and `foundFilename` kept refusing to be true (for example, because of some permissions thing), I think it's unlikely we should try more than, say, 10 different filenames. (We certainly wouldn't want to be trying millions e.g. `/tmp/secret21674364736743.txt`)

So I propose to use the simpler `for` form: `for suffix := 1;  suffix<=20; suffix++ {` and have the for loop return on success. If it falls out of the loop, we know something weird happened, and can return an error.

fk/secretreceive.go
paulfurley: I saw `checker` and thought ... hmm... it must be a "file exists" function (then discovered it meant "file *doesn't* exist).

I think we could avoid this confusion with e.g. `fileExistsInterface`, and that's the Go norm.

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+		doesntExist,
+	)
+
+	return filenameToWrite, nil
+}
+
+func getNewUniqueFilename(fp string, checker func(string) bool) string {
+	i := 0
+	foundFilename := false
+	filename := filepath.Base(fp)
```
paulfurley: Again, is `filename` the *base* filename or the *full filename* ?

fk/secretreceive_test.go
```diff
@@ -194,5 +203,73 @@ func TestDecryptAPISecret(t *testing.T) {
...
+			assert.Equal(t, decryptedSecret.UUID, uuid)
+		})
+
+		t.Run("with a matching filename", func(t *testing.T) {
+			assert.Equal(t, decryptedSecret.filename, "example.txt")
+		})
+	})
+}
+
+func mockDoesntExist(path string) bool {
```
paulfurley: Ha! I see! Well... that works... it's a divergence from the interface pattern though!

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+	numberedFilename := filename
+	for ok := true; ok; ok = (foundFilename == false) {
+		if i == 0 {
+			// try using the original filename it was sent as first with no count suffix
+			numberedFilename = filename
+		} else if i > 0 {
+			// the original filename already exists, try `filename(i).ext` next...
+			numberedFilename = fmt.Sprintf(
```
paulfurley: I'd propose `/tmp/secret.1.txt` beats `/tmp/secret(1).txt`

fk/secretreceive_test.go
```diff
@@ -194,5 +203,73 @@ func TestDecryptAPISecret(t *testing.T) {
...
+	}
+
+	if path == "/fake/old_filename.txt.bak" {
+		return false
+	}
+
+	return true
+}
+
+func TestGetNewUniqueFilename(t *testing.T) {
```
paulfurley: Recommend adding a test "gives up if it can't find a filename after 10 attempts"

fk/secretreceive.go
paulfurley: actually maybe not.

fk/secretreceive.go
```diff
@@ -194,6 +198,10 @@ func decryptAPISecret(
 		UUID:             uuid,
 	}
 
+	if !literalData.ForEyesOnly() {
+		decryptedSecret.filename = literalData.FileName
```
paulfurley: So what happens if I encode a secret with the filename set to a full path like `/Users/ian/.zsh` ?

fk/secretreceive.go
```diff
@@ -202,13 +230,77 @@ func countDigits(i int) (count int) {
...
+		if os.IsNotExist(err) {
+			return "", fmt.Errorf(downloadsDir + " directory doesn't exist")
+		} else {
+			return "", err
+		}
+	}
+	if !fileInfo.IsDir() {
+		return "", fmt.Errorf(fileInfo.Name() + " file exists and is not a directory")
+	}
+	filenameToWrite := getNewUniqueFilename(
```
paulfurley: All that business with the `Downloads` directory and it seems this returns a basename ... e.g.

```
getNewUniqueFilename("/tmp/secret.txt") -> `secret.txt`
```

fk/secretreceive.go
idrysdale: Yeah, *maybe*, I think seeing "~/Downloads" would help actually... a reminder where....

fk/secretreceive.go
idrysdale: Whoops! Yep, I came across this concern in writing this, but missed this one! Thanks!

fk/secretreceive.go
idrysdale: 👍

fk/secretreceive.go
idrysdale: 👍 :)

fk/secretreceive.go
idrysdale: .... and again....

fk/secretreceive.go
idrysdale: Ok!

fk/secretreceive.go
idrysdale: Sensible.

fk/secretreceive.go
idrysdale: Yeah, I dunno - I followed FireFox's example!

fk/secretreceive_test.go
idrysdale: 😝

fk/secretreceive.go
paulfurley: Yeah, to remind, but also to *review* whether to allow it: I would assume `secret.txt` meant `~/secret.txt`, and would say "no".

(Also, an aside, it doesn't actually save to `~/Downloads` even if you say yes)

idrysdale: Sorry - we're agreeing!

fk/secretreceive.go
```diff
@@ -84,12 +91,26 @@ func secretReceive() exitCode {
...
-				err := clipboard.WriteAll(secret.decryptedContent)
+			out.Print(formatSecretListItem(secret.decryptedContent, secret.filename))
+			if secret.filename != "" {
+				filenameToWrite, err := getFileNameToWrite(secret.filename)
 				if err != nil {
+					printFailed("Error prompting to save file to disk:")
 					printFailed(err.Error())
 				}
+				if prompter.promptYesNo("Save to "+filenameToWrite+"?", "", nil) == true {
+					err := ioutil.WriteFile(filenameToWrite, []byte(secret.decryptedContent), 0644)
```
paulfurley: `0644` ? :face_with_head_bandage:

fk/secretreceive.go
paulfurley: I ended up setting on `IsSafeToWrite` which is more explicit (it's not actually `!fileExists` because what if it exists AND it cannot be written?)

fk/secretreceive.go
paulfurley: See https://github.com/fluidkeys/fluidkeys/pull/294/commits/f5b10cbf602c79613655a6187d5d4d4020bc357d

PR #295, opened 2019-02-06
factor out downloadEncryptedSecrets and decryptSecrets

idrysdale:
We've factored `downloadAndDecryptSecrets` into `downloadEncryptedSecrets` and `decryptSecrets`.

We've added tests for `decryptAPISecret` and `downloadEncryptedSecrets`.


paulfurley:
I'm looking at this now :)

PR #298, opened 2019-02-07
check file isn't binary and doesn't contain bad runes *before* previewing it to send

idrysdale:
I think it's worth doing it here *and* if the user goes on to confirm
they want to send the file at prompt, as in theory they could
maliciously overwrite the file after the first check.

### DIFF
--- a/fk/secretsend.go
+++ b/fk/secretsend.go
@@ -126,6 +126,13 @@ func getSecretFromFile(filename string, fileReader ioutilReadFileInterface, prom
 	if len(strings.TrimSpace(secret)) == 0 {
 		return "", fmt.Errorf(filename + " is empty")
 	}
+	if !utf8.ValidString(secret) {
+		return "", fmt.Errorf(filename + " is not valid utf8 (is it a binary?)")
+	}
+	if stringutils.ContainsDisallowedRune(secret) {
+		return "", fmt.Errorf(filename + " contained disallowed characters")
+	}
+
 	out.Print("---\n")
 	out.Print(secret)
 	out.Print("---\n\n")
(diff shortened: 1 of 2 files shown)


fk/secretsend.go
```diff
@@ -126,6 +126,13 @@ func getSecretFromFile(filename string, fileReader ioutilReadFileInterface, prom
 	if len(strings.TrimSpace(secret)) == 0 {
 		return "", fmt.Errorf(filename + " is empty")
 	}
+	if !utf8.ValidString(secret) {
```
paulfurley: Ooh GOOD FIND!

fk/secretsend_test.go
```diff
@@ -109,6 +109,28 @@ func TestGetSecretFromFile(t *testing.T) {
 		assert.Equal(t, expectedErr, err)
 	})
 
+	t.Run("returns error if file is binary", func(t *testing.T) {
```
paulfurley: FYI *all* files are binary, I think the test really means "returns error if file won't decode as UTF-8"

fk/secretsend.go
paulfurley: I *suspect* that `utf8.ValidString` probably covers all of our disallowed rune cases... but never mind!

fk/secretsend.go
```diff
@@ -126,6 +126,13 @@ func getSecretFromFile(filename string, fileReader ioutilReadFileInterface, prom
...
+		return "", fmt.Errorf(filename + " is not valid utf8 (is it a binary?)")
+	}
+	if stringutils.ContainsDisallowedRune(secret) {
```
paulfurley: So why not put the `utf8.ValidString` *inside* our validator `ContainsDisallowedRune` ?

fk/secretsend.go
paulfurley: You're right, it can be utf8 *and* have disallowed runes

PR #299, opened 2019-02-07
block invalid runes on DecryptArmoredToString

idrysdale:


paulfurley:
I don't understand... why?

PR #302, opened 2019-02-08
dont panic if matching email for key exists on api

idrysdale:
screenshot 2019-02-08 at 11 13 07

### DIFF
--- a/fk/keycreate.go
+++ b/fk/keycreate.go
@@ -139,7 +139,9 @@ func keyCreate(email string) (exitCode, *pgpkey.PgpKey) {
 			verifiedEmailResult, err = verifyEmailMatchesKeyInAPI(
 				email, generateJob.pgpKey.Fingerprint(), client)
 			if err != nil {
-				log.Panicf("Failed to get public key: %v", err)
+				out.Print("\n\n")
+				printFailed(fmt.Sprintf("Failed to verify email: %s", err))
+				return 1, nil
 			}
 			timeLastPolled = time.Now()
 		}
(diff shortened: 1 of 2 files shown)


fk/keycreate.go
```diff
@@ -248,11 +250,18 @@ func verifyEmailMatchesKeyInAPI(
 
 	retrievedKey, err := pgpkey.LoadFromArmoredPublicKey(armoredKey)
 	if err != nil {
-		return false, fmt.Errorf("Failed to load armored key: %v", err)
+		return false, fmt.Errorf("failed to load armored key: %v", err)
 	}
 	if retrievedKey.Fingerprint() == fingerprint {
 		return true, nil
 	}
+	retrievedEmail, err := retrievedKey.Email()
```
paulfurley: This *new* test is cute: ensure the API didn't return as a duff key which actually has no email addresses on it.

But did you mean to introduce it? Are there other consequences?

fk/keycreate.go
```diff
@@ -248,11 +250,18 @@ func verifyEmailMatchesKeyInAPI(
...
+	if err != nil {
+		return false, fmt.Errorf("couldn't get email for retrieved key: %s", err)
+	}
+	if retrievedEmail == email {
```
paulfurley: Unfortunately this logic isn't correct: if the key has multiple email addresses, `Email()` chooses the most "primary" one. So it's perfectly valid for `retrievedEmail` not to match `

fk/keycreate.go
```diff
@@ -248,11 +250,18 @@ func verifyEmailMatchesKeyInAPI(
...
+		return false, fmt.Errorf("a key for %s is already verified\n     Please email security@fluidkeys.com and we can manually remove the old key\n", retrievedEmail)
+	}
 	return false, fmt.Errorf("Retrieved key's fingerprint doesn't match")
```
paulfurley: *this* is the line you probably meant to replace... right?

I'd suggest flipping the logic first so the failure case comes first, and the success case comes last (I did this on a WIP branch: 69b96db2b47337e9e27d075bb35366b9664f807d)

paulfurley:
Perfect!

PR #304, opened 2019-02-08
automatically delete secrets and warn user

idrysdale:


fk/secretreceive.go
```diff
@@ -110,14 +111,16 @@ func secretReceive() exitCode {
...
-				}
+
+			err := client.DeleteSecret(key.Fingerprint(), secret.UUID.String())
+			if err != nil {
+				log.Printf("failed to delete secret '%s': %v", secret.UUID, err)
+				printFailed("Error when secret tried to self-destruct:")
+				printFailed(err.Error())
 			}
+
+			deletedSecret = true
```
paulfurley: This is set even if there's an error deleting a secret, in which case "Secrets have self destructed!" would still be printed

paulfurley: (Although I guess it has already printed the failure message at that point, so it's OK)

fk/secretreceive.go
```diff
@@ -130,6 +133,10 @@ func secretReceive() exitCode {
 		}
 	}
 
+	if deletedSecret {
```
paulfurley: Currently `if deletedSecret` could be replaced with `if len(decryptedSecrets)`... e.g. it's redundant.

Either `deletedSecrets` should actually mean *deleted secrets **without error*** or it should disappear (variables are bad)

fk/secretreceive.go
```diff
@@ -130,6 +133,12 @@ func secretReceive() exitCode {
 		}
 	}
 
+	if numSecretsDeleted > 0 {
+		deleteMessage := humanize.Pluralize(numSecretsDeleted, "secret", "secrets") +
```
paulfurley: not quite: "💥 1 secret **have** self destructed!"

PR #307, opened 2019-02-08
Improve the user interface on `secret send`

idrysdale:
screenshot 2019-02-08 at 18 08 59

### DIFF
--- a/fk/secretsend.go
+++ b/fk/secretsend.go
@@ -77,9 +77,9 @@ https://download.fluidkeys.com#` + recipientEmail + `
 	if filename != "" {
 		secret, err = getSecretFromFile(filename, nil)
 
-		printFileDivider(filename)
+		out.Print(formatFileDivider(filename) + "\n")
 		out.Print(secret)
-		printFileDivider("")
+		out.Print(formatFileDivider("") + "\n")
 		out.Print("\n")
 
 		out.Print(colour.Info("The file will be end-to-end encrypted to ") + recipientEmail + "\n")
--- a/fk/secretsend.go
+++ b/fk/secretsend.go
@@ -81,7 +81,9 @@ https://download.fluidkeys.com#` + recipientEmail + `
 			return 1
 		}
 
-		out.Print(formatFileDivider(filename) + "\n")
+		basename = filepath.Base(filename)
+
+		out.Print(formatFileDivider(basename) + "\n")
 		out.Print(secret)
 		out.Print(formatFileDivider("") + "\n")
 		out.Print("\n")
@@ -91,11 +93,9 @@ https://download.fluidkeys.com#` + recipientEmail + `
 
 		prompter := interactiveYesNoPrompter{}
 
-		if !prompter.promptYesNo("Send "+filename+"?", "y", nil) {
+		if !prompter.promptYesNo("Send "+basename+"?", "y", nil) {
 			return 1
 		}
-
-		basename = filepath.Base(filename)
 	} else {
 		out.Print(colour.Info("Type or paste your message, ending by typing Ctrl-D\n"))
 		out.Print(colour.Info("It will be end-to-end encrypted to ") + recipientEmail + "\n")
--- a/fk/secretsend.go
+++ b/fk/secretsend.go
@@ -58,8 +58,6 @@ https://download.fluidkeys.com#` + recipientEmail + `
 		return 1
 	}
 
-	printSuccess("Found public key for " + recipientEmail)
-
 	pgpKey, err := pgpkey.LoadFromArmoredPublicKey(armoredPublicKey)
 	if err != nil {
 		printFailed("Couldn't load the public key:")
@@ -84,7 +82,8 @@ https://download.fluidkeys.com#` + recipientEmail + `
 		printFileDivider("")
 		out.Print("\n")
 
-		out.Print(colour.Info("   It will be end-to-end encrypted so no-one else can read it\n\n"))
+		out.Print(colour.Info("The file will be end-to-end encrypted to ") + recipientEmail + "\n")
+		out.Print(colour.Info("so no-one else can read it 🕵️\n\n"))
 
 		prompter := interactiveYesNoPrompter{}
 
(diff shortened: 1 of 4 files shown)


paulfurley:
Didya run the tests? :)

fk/secretsend.go
```diff
@@ -77,9 +77,25 @@ https://download.fluidkeys.com#` + recipientEmail + `
...
+
+		out.Print("---\n")
+		out.Print(secret)
+		out.Print("---\n\n")
+
+		out.Print(colour.Info("   It will be end-to-end encrypted so no-one else can read it\n\n"))
+
+		prompter := interactiveYesNoPrompter{}
+
+		if !prompter.promptYesNo("Send "+filename+"?", "y", nil) {
```
paulfurley: Cool, to confirm, it now doesn't print an error message, it just quits. That seems good to me.

idrysdale: That's right.

fk/secretsend_test.go
```diff
@@ -128,27 +116,11 @@ func TestGetSecretFromFile(t *testing.T) {
 			readFileBytes: []byte(""),
 		}
 
-		_, err := getSecretFromFile("/fake/filename", fileReader, nil)
+		_, err := getSecretFromFile("/fake/filename", fileReader)
 		expectedErr := fmt.Errorf("/fake/filename is empty")
 		assert.Equal(t, expectedErr, err)
 	})
 
-	t.Run("returns error if user answers no", func(t *testing.T) {
```
paulfurley: 'tis a shame to delete this test and not add it back in at a higher level: this particular test was introduced because at one point we hit a nil pointer error if the user said no. It was only by luck that we found that error while messing around on the CLI.

fk/secretsend.go
```diff
@@ -118,7 +118,7 @@ https://download.fluidkeys.com#` + recipientEmail + `
 		return 1
 	}
 
-	printSuccess("Successfully sent secret to " + recipientEmail + "\n")
```
paulfurley: :+1:

fk/ui.go
```diff
@@ -51,3 +52,17 @@ func printFailedAction(message string) {
 func printHeader(message string) {
 	out.Print(colour.Header(fmt.Sprintf(" %-79s", message)) + "\n\n")
 }
+
+func printFileDivider(message string) {
```
paulfurley: It would be awesome to add the example from the commit message to a comment on this function. From the code I struggled to work out what it's for!
e.g.
```
// printFileDivider prints a divider string, including a message if given, e.g.:
// ── readme.txt ──────────────────────────────────────────────────
```

fk/ui.go
```diff
@@ -51,3 +52,17 @@ func printFailedAction(message string) {
...
+	if message != "" {
+		message = " " + message + " "
+	}
+	leftHandAndMessage := strings.Repeat(fileDividerRune, 2) + colour.File(message)
```
paulfurley: Does the fact that `message` is always `colour.File(..)` imply that `message` should actually be called `filename` ?

fk/ui.go
```diff
@@ -51,3 +52,17 @@ func printFailedAction(message string) {
...
+	rightHand := strings.Repeat(fileDividerRune, (fileDividerLength - len(leftHandAndMessage)))
```
paulfurley: Ooh I would have expected `len(leftHandMessage)` to include the length of the `colour.File(..)` control bytes, is that not the case

fk/ui.go
paulfurley: I've just pushed a commit to add this!

fk/secretsend.go
```diff
@@ -79,9 +79,10 @@ https://download.fluidkeys.com#` + recipientEmail + `
 	if filename != "" {
 		secret, err = getSecretFromFile(filename, nil)
 
-		out.Print("---\n")
+		printFileDivider(filename)
```
paulfurley: I suspect this should be the *basename* e.g. `aws-cert.pem` rather than `~/.ssh/aws-cert.pem` ... since the information we're transmitting does not include the full path.

paulfurley:
One more thing: a long filename causes `printFileDivider` to crash:
```
$ ffk secret send /tmp/fedhufndsaujfnjudsafbdsaukfbheufndsajcndsaunfdjhukanfdujkanfujdanfuwenfuejdsanfuidsanfudjsdusnuianuifdeas.txt --to paul@fluidkeys.com
go build -o build/bin/fk cmd/fk/main.go
panic: strings: negative Repeat count

goroutine 1 [running]:
strings.Repeat(0x84a162, 0x3, 0xffffffffffffffcd, 0xc42043a300, 0x7d)
        /usr/lib/go-1.10/src/strings/strings.go:538 +0x1c7
github.com/fluidkeys/fluidkeys/fk.printFileDivider(0x7ffe031b4f33, 0x72)
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/fk/ui.go:63 +0x158
github.com/fluidkeys/fluidkeys/fk.secretSend(0x7ffe031b4fab, 0x12, 0x7ffe031b4f33, 0x72, 0x72)
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/fk/secretsend.go:80 +0x444
github.com/fluidkeys/fluidkeys/fk.secretSubcommand(0xc4203ec300, 0xc4203ed380)
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/fk/main.go:322 +0x29c
github.com/fluidkeys/fluidkeys/fk.Main(0xc42009c058)
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/fk/main.go:96 +0x35d
main.main()
        /home/paul/go/src/github.com/fluidkeys/fluidkeys/cmd/fk/main.go:26 +0x22

```

fk/ui.go
idrysdale: I don't follow? I think it does?

fk/ui.go
idrysdale: Yep! Good spot!

fk/secretsend.go
```diff
@@ -81,7 +81,9 @@ https://download.fluidkeys.com#` + recipientEmail + `
 			return 1
 		}
 
-		out.Print(formatFileDivider(filename) + "\n")
+		basename = filepath.Base(filename)
```
paulfurley: :+1:

PR #324, opened 2019-02-18
stop passing time into gpg.BackupHomeDir

idrysdale:
This code is a hangover from when we used to timestamp the file backup. We don't do this anymore since we put the file in a timestamped directory (as seen in `makeGnupgBackup`).

Comment `gpgwrapper.Backuphomedir` to make this clear.


paulfurley:
Nice!

PR #328, opened 2019-02-19
don't panic on cron error (print pretty message)

paulfurley:
* add new package ui: FormatFailure, FormatWarning
* improve crontab error messages on writeCrontab
* extract runCrontab into interface for mocking later

### DIFF
--- a/scheduler/interfaces.go
+++ b/scheduler/interfaces.go
@@ -1,6 +1,6 @@
 package scheduler
 
 type runCrontabInterface interface {
-	runCrontab(arguments ...string) (string, error)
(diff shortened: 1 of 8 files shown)


paulfurley:
image

scheduler/interfaces.go
```diff
@@ -0,0 +1,6 @@
+package scheduler
+
+type runCrontabInterface interface {
```
idrysdale: IIRC, we created this interface was to allow us to mock it in tests, but we haven't. I suggest we either write some tests that mock it, or remove this interface from this PR.

PR #332, opened 2019-02-20
team setup: prompt for team name and members

idrysdale:
This command doesn't actually do anything with the information it collects
at this stage. It just prints it back to screen and warns the user that
this feature isn't currently implemented.

### DIFF
--- a/fk/main.go
+++ b/fk/main.go
@@ -66,6 +66,7 @@ Configuration file: %s
 Usage:
 	fk setup
 	fk setup <email>
+	fk team setup
 	fk team sync [--cron-output]
 	fk secret send <recipient-email>
 	fk secret send [<filename>] --to=<email>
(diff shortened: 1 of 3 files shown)


fk/teamcreate_test.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/fluidkeys/fluidkeys/colour"
+
+	"github.com/fluidkeys/fluidkeys/assert"
+)
+
+func TestValidateTeamName(t *testing.T) {
```
paulfurley: nice!!

fk/main.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/fluidkeys/fluidkeys/api"
 	"github.com/fluidkeys/fluidkeys/fingerprint"
 
-	"github.com/docopt/docopt-go"
+	docopt "github.com/docopt/docopt-go"
```
paulfurley: what's this change about?

idrysdale: my editor wants to refence this pacakage as `docopt` when I save. I guess because the pacakge filename is `docopt-go` but I'm not sure 🤷‍♂️ Anyway, I've dropped the commit from this PR.

PR #338, opened 2019-02-22
add ValidateRoster to Team

idrysdale:


teamroster/teamroster.go
```diff
@@ -34,6 +34,31 @@ func Load(fluidkeysDirectory string) ([]Team, error) {
 	return teams, nil
 }
 
+// ValidateRoster asserts that the team roster has no email addresses or fingerprints that are
+// listed more than once.
+func (t *Team) ValidateRoster() error {
```
paulfurley: Ah interesting adding it to `Team`, I expected you to add it to `teamroster.Valid` ... but on `Team` seems good.

It raises a question about whether our `Team` is named correctly. I'd expect to do `<the roster>.Validate()` rather than `<the roster>.ValidateRoster()` but since the `Team` **is** the roster that would look like `team.Validate()`

paulfurley: Thinking more about it based on how I use it in the API, I'd prefer it was internally part of `Parse`, so I can just do:
```
 	roster, err := teamroster.Parse(strings.NewReader(requestData.TeamRoster))
        <syntax *and* logic errors are included>
```
```

idrysdale: Yeah - that makes sense. Let's move it there!

PR #346, opened 2019-02-26
Team Create: Upload roster and signature to server

idrysdale:
Finally, during `team create` we can now upload the roster and signature to the server.

This is very much and MVP. It's not really handling errors correctly (i.e. we always return `0`, regardless of whether we encountered a problem along the way, like uploading to server).

screenshot 2019-02-26 at 19 25 11


paulfurley:
I'm rebasing now

PR #347, opened 2019-02-26
Add padding to account for checkbox when printing roster path

idrysdale:
screenshot 2019-02-26 at 19 30 56

### DIFF
--- a/fk/teamcreate.go
+++ b/fk/teamcreate.go
@@ -176,7 +176,7 @@ func teamCreate() exitCode {
 		printCheckboxFailure("Create and sign team roster", err)
 	}
 	printCheckboxSuccess("Create and sign team roster in \n" +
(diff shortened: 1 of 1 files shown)


fk/teamcreate.go
```diff
@@ -176,7 +176,7 @@ func teamCreate() exitCode {
 		printCheckboxFailure("Create and sign team roster", err)
 	}
 	printCheckboxSuccess("Create and sign team roster in \n" +
-		filepath.Join(fluidkeysDirectory, "teams"))
+		"         " + filepath.Join(fluidkeysDirectory, "teams")) // account for checkbox indent
```
paulfurley: I've added a tech debt card to provide a second argument to `printCheckboxSuccess` to take a second `lines` argument since we use this pattern quite a few times.

PR #348, opened 2019-02-26
allow Team.SignAndSave to overwrite the files

idrysdale:


team/team.go
```diff
@@ -47,9 +47,11 @@ func SignAndSave(team Team, fluidkeysDirectory string, signingKey *pgpkey.PgpKey
 		getTeamDirectory(fluidkeysDirectory), // ~/.config/fluidkeys/teams
 		team.subDirectory(),                  // fluidkeys-inc-4367436743
 	)
-	if _, err := os.Stat(rosterDirectory); !os.IsNotExist(err) {
-		return fmt.Errorf("path already exists at %s", rosterDirectory)
+	_, err = os.Stat(rosterDirectory)
```
paulfurley: What’s the purpose of calling `os.Statm”` at all on a directory we expect not to exist?

idrysdale: I figured it may exist - like if you call `Save` a second time, in which case we don't want to make it!

PR #358, opened 2019-02-28
rename ErrorIsNil > NoError

idrysdale:


paulfurley:
what's the difference between this and #357? or did we independently just implement this same thing?!

PR #362, opened 2019-02-28
add client.GetTeamName

idrysdale:


api/api.go
```diff
@@ -55,6 +55,9 @@ type Client struct {
 // ErrPublicKeyNotFound means the response was OK, but no key was found
 var ErrPublicKeyNotFound = fmt.Errorf("Public key not found")
 
+// ErrTeamNotFound means the response was OK, but no team was found
+var ErrTeamNotFound = fmt.Errorf("Team not found")
```
paulfurley: cool!

api/api_test.go
```diff
@@ -332,6 +334,80 @@ func TestUpsertTeam(t *testing.T) {
...
+func TestGetTeamName(t *testing.T) {
+	t.Run("parses the name from a good response", func(t *testing.T) {
+		client, mux, _, teardown := setup()
+		defer teardown()
+
+		teamUUID := uuid.Must(uuid.NewV4())
+		teamResponse, err := json.Marshal(v1structs.GetTeamResponse{
+			Name: "Kiffix Ltd",
+		})
+		if err != nil {
```
paulfurley: i find `assert.NoError` more helpful in cases like this one, since it's 1 line rather than 3, and it doesn't swallow the content of the error

api/api_test.go
```diff
@@ -332,6 +334,80 @@ func TestUpsertTeam(t *testing.T) {
...
+			fmt.Sprintf("/team/%s", unknownUUID),
+			mockResponseHandler,
+		)
+
+		_, err := client.GetTeamName(unknownUUID)
+
+		assert.Equal(t, ErrTeamNotFound, err)
+	})
+
+	t.Run("responds with http 500 (unexpected http code)", func(t *testing.T) {
```
paulfurley: the name might be a littler clearer if it was "client returns error if API returns HTTP 500"

api/api_test.go
```diff
@@ -332,6 +334,80 @@ func TestUpsertTeam(t *testing.T) {
...
+		}
+		mux.HandleFunc(
+			fmt.Sprintf("/team/%s", teamUUID),
+			mockResponseHandler,
+		)
+
+		_, err := client.GetTeamName(teamUUID)
+
+		assert.GotError(t, err)
+		assert.Equal(t, fmt.Errorf("API error: 500"), err)
```
paulfurley: Not sure if this is a comment on this PR in particular, but the API *always* returns a `{"detail": "some more information"}` JSON structure if it hits an error. I'd expect the API client to pass these through

paulfurley: I *think* it does pass it through, but perhaps it's because the `mockResponseHandler` isn't creating that error detail

PR #367, opened 2019-02-28
Refer to the fingerprint package as fpr and value as fingerprint

idrysdale:
We should always refer to the `fingerprint` package as `fpr`.


paulfurley:
Would be great to get this in (requiring a rebase) or close it

idrysdale:
Rebased! Ready for review 🤘🏼

PR #368, opened 2019-03-04
Add client.ListRequestsToJoinTeam

idrysdale:


fingerprint/fingerprint_test.go
```diff
@@ -39,6 +39,11 @@ func TestFingerprint(t *testing.T) {
 			"A999 B749 8D1A 8DC4 73E5  3C92 309F 635D AD1B 5517",
 			nil, // good fingerprint: no error
 		},
+		{
```
paulfurley: Nice!! it's valid for `OPENPGP4FPR` to be lowercase, so we need a test case for this too

idrysdale: 👍

api/api.go
```diff
@@ -274,6 +273,41 @@ func (c *Client) RequestToJoinTeam(
...
+	_, err = c.do(request, &decodedJSON)
+	if err != nil {
+		return nil, err
+	}
+	for _, jsonRequestToJoin := range decodedJSON.Requests {
+		requestUUID, err := uuid.FromString(jsonRequestToJoin.UUID)
+		if err != nil {
+			continue
+		}
+		requestFingerprint, err := fpr.Parse(jsonRequestToJoin.Fingerprint)
```
paulfurley: One day I'd like to change the `v1structs` so that fingerprint is a `fpr.Fingerprint` rather than a `string`, then all of this parsing and formatting will just work automatically... imagine!

paulfurley:
Nice one!

PR #369, opened 2019-03-04
record administrators in Team

idrysdale:
We'll need to bump the package on `fluidkeys/api` too, so it can parse the incoming toml file.

### DIFF
--- a/fk/teamcreate.go
+++ b/fk/teamcreate.go
@@ -49,7 +49,7 @@ func teamCreate() exitCode {
 		return 1
 	}
 
-	teamMembers := []team.Person{{Email: email, Fingerprint: key.Fingerprint()}}
+	teamMembers := []team.Person{{Email: email, Fingerprint: key.Fingerprint(), IsAdmin: true}}
 
 	printHeader("What's your team name?")
 
--- a/team/roster_test.go
+++ b/team/roster_test.go
@@ -13,6 +13,27 @@ import (
 )
 
 func TestParse(t *testing.T) {
+
+	const validRoster = `# Fluidkeys team roster
+
+uuid = "38be2a70-23d8-11e9-bafd-7f97f2e239a3"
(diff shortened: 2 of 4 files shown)


paulfurley:
... and possible validate the team's admin differently, too

team/roster_test.go
```diff
@@ -60,10 +61,12 @@ name = "Kiffix"
 [[person]]
   email = "test2@example.com"
   fingerprint = "5C78E71F6FEFB55829654CC5343CC240D350C30C"
+  is_admin = true
```
paulfurley: it would be *super* to see a test case where `is_admin` is *not* set (and that it defaults to `false`!!)

team/team.go
```diff
@@ -192,4 +191,5 @@ func (t *Team) Fingerprints() []fpr.Fingerprint {
 type Person struct {
 	Email       string          `toml:"email"`
 	Fingerprint fpr.Fingerprint `toml:"fingerprint"`
+	IsAdmin     bool            `toml:"is_admin"`
```
paulfurley: Do you know if it's possible to set defaults in these annotations?

team/roster_test.go
paulfurley: Well, question, maybe it's not OK to have `is_admin` missing? And we just trash all instances of existing rosters? But in the future we'll have to support adding new fields with graceful defaults.

fk/teamcreate.go
```diff
@@ -49,7 +49,7 @@ func teamCreate() exitCode {
 		return 1
 	}
 
-	teamMembers := []team.Person{{Email: email, Fingerprint: key.Fingerprint()}}
+	teamMembers := []team.Person{{Email: email, Fingerprint: key.Fingerprint(), IsAdmin: true}}
```
paulfurley: :+1:

PR #373, opened 2019-03-12
Start making database more general

idrysdale:
This is in preparation for using it to also record which teams
a user has requested to join.

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -77,9 +77,8 @@ func (db *Database) GetFingerprintsImportedIntoGnuPG() ([]fpr.Fingerprint, error
 	if err != nil {
 		if os.IsNotExist(err) {
 			return []fpr.Fingerprint{}, nil
-		} else {
-			return nil, fmt.Errorf("Couldn't open '%s': %v", db.jsonFilename, err)
 		}
+		return nil, fmt.Errorf("Couldn't open '%s': %v", db.jsonFilename, err)
 	}
 
 	byteValue, err := ioutil.ReadAll(file)
@@ -122,7 +121,7 @@ func deduplicate(slice []fpr.Fingerprint) []fpr.Fingerprint {
 	}
 
 	var deduped []fpr.Fingerprint
-	for key, _ := range sliceMap {
+	for key := range sliceMap {
 		deduped = append(deduped, key)
 	}
 	return deduped
--- a/database/database.go
+++ b/database/database.go
@@ -27,23 +27,29 @@ import (
 	fpr "github.com/fluidkeys/fluidkeys/fingerprint"
 )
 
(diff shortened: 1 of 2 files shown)


database/database.go
```diff
@@ -31,7 +31,7 @@ type Database struct {
 	jsonFilename string
 }
 
-type DatabaseMessage struct {
+type Message struct {
```
paulfurley: Agree that renaming the top level struct to `Message` is better because of `database.DatabaseMessage.

database/database.go
```diff
@@ -51,7 +51,7 @@ func (db *Database) RecordFingerprintImportedIntoGnuPG(newFingerprint fpr.Finger
 	}
 
 	allFingerprints := append(existingFingerprints, newFingerprint)
-	databaseMessage := makeDatabaseMessageFromFingerprints(deduplicate(allFingerprints))
+	message := makeMessageFromFingerprints(deduplicate(allFingerprints))
```
paulfurley: I feel renaming all these internal instances of `databaseMessage` to just `message` have made it less clear what's happening. First impressions are that it's now tricky to grok whether we're looking at a `databaseMessage` or one of its sub messages.

paulfurley: Come to think of it it feels pretty strange referring to it as a *message* at all, where it's actually a *database*

database/database.go
```diff
@@ -51,15 +51,20 @@ func New(fluidkeysDirectory string) Database {
 // RecordFingerprintImportedIntoGnuPG takes a given fingperprint and records that it's been
 // imported into GnuPG by writing an updated json database.
 func (db *Database) RecordFingerprintImportedIntoGnuPG(newFingerprint fpr.Fingerprint) error {
```
paulfurley: I'm struggling to work out what's happening in this commit: there seems to be quite a lot of changes, and the commit message doesn't give me many clues!

database/database.go
```diff
@@ -65,22 +65,22 @@ func (db *Database) RecordFingerprintImportedIntoGnuPG(newFingerprint fpr.Finger
...
-	for _, v := range databaseMessage.KeysImportedIntoGnuPG {
+	for _, v := range message.KeysImportedIntoGnuPG {
 		fingerprints = append(fingerprints, v.Fingerprint)
 	}
 
 	return deduplicate(fingerprints), nil
 }
 
-func (db *Database) getMessage() (message *Message, err error) {
+func (db *Database) readMessage() (message *Message, err error) {
```
paulfurley: I'm gonna suggest we go really explicit and call these
* `readMessage` -> `loadFromFile`
* `writeMessage` -> `saveToFile`

PR #375, opened 2019-03-12
Record in db when user requests to join a team

idrysdale:


fk/teamjoin.go
```diff
@@ -60,6 +61,12 @@ func teamJoin(teamUUID uuid.UUID) exitCode {
 	printCheckboxSuccess(action)
 	out.Print("\n\n")
 
+	db.RecordRequestToJoinTeam(team.RequestToJoinTeam{
+		UUID:        teamUUID,
```
paulfurley: nope, `UUID` is the **request** UUID. This has tripped us up a few times so I propose we make it explicit by renaming `UUID` to `RequestUUID`

fk/teamjoin.go
```diff
@@ -60,6 +61,12 @@ func teamJoin(teamUUID uuid.UUID) exitCode {
 	printCheckboxSuccess(action)
 	out.Print("\n\n")
 
+	db.RecordRequestToJoinTeam(team.RequestToJoinTeam{
```
paulfurley: We don't have a timestamp, so `fk team sync` won't know when to give up!

PR #387, opened 2019-03-15
Fix missing return error

paulfurley:


idrysdale:
I'm assuming that this means you'll tackle error handling differently to how I started on `team sync` here:
https://github.com/fluidkeys/fluidkeys/blob/wip-get-requests/fk/teamsync.go#L36 ?

paulfurley:
Sure, anyone calling this will have to actually look at the error :) But it shouldn't be swallowing errors regardless of what callers do with them.

idrysdale:
Ok, but this also swallows it in the same way:
https://github.com/fluidkeys/fluidkeys/blob/master/database/database.go#L109

I think the rationale was that if the db hasn't yet been written when performing the get, return an empty slice and swallow error. This assumes that we'd then write the file out as part of 'save'.

paulfurley:
OHHH it's pretending it's an empty list. I gotcha

idrysdale:
👍

paulfurley:
Converted into tech debt card to think about some other time: https://trello.com/c/gf9wdQn5

PR #396, opened 2019-03-18
api.GetTeamRoster: 403: return ErrForbidden

paulfurley:
this is so that `fk team sync` can handle it as "you don't have access to
this team"

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -493,6 +493,23 @@ func TestGetTeamRoster(t *testing.T) {
 		assert.Equal(t, ErrTeamNotFound, err)
 	})
 
+	t.Run("403 forbidden returns ErrForbidden", func(t *testing.T) {
+		teamUUID := uuid.Must(uuid.NewV4())
+		mockResponseHandler := func(w http.ResponseWriter, r *http.Request) {
+			assertClientSentVerb(t, "GET", r.Method)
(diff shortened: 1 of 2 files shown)


api/api_test.go
```diff
@@ -493,6 +493,23 @@ func TestGetTeamRoster(t *testing.T) {
 		assert.Equal(t, ErrTeamNotFound, err)
 	})
 
+	t.Run("403 forbidden returns ErrForbidden", func(t *testing.T) {
+		unknownUUID := uuid.Must(uuid.NewV4())
+		mockNotFoundResponseHandler := func(w http.ResponseWriter, r *http.Request) {
```
idrysdale: Ditto, I think this should probably be called 'mockForbiddenResponseHandler`

api/api_test.go
```diff
@@ -493,6 +493,23 @@ func TestGetTeamRoster(t *testing.T) {
 		assert.Equal(t, ErrTeamNotFound, err)
 	})
 
+	t.Run("403 forbidden returns ErrForbidden", func(t *testing.T) {
+		unknownUUID := uuid.Must(uuid.NewV4())
```
idrysdale: Maybe this should be called the more generic `uuid` or `forbiddenUUID`?

paulfurley:
Good spot!

PR #397, opened 2019-03-18
Add database.GetRequestsToJoinTeam(uuid)

idrysdale:
This will be useful to check whether an existing request to join a team already exists.

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -163,6 +163,24 @@ func (db *Database) DeleteRequestToJoinTeam(teamUUID uuid.UUID, fingerprint fpr.
 	return db.saveToFile(*message)
 }
 
+// GetRequestsToJoinTeam returns a slice of requests to join the team matching the UUID.
+func (db *Database) GetRequestsToJoinTeam(teamUUID uuid.UUID) (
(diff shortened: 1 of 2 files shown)


paulfurley:
Needs rebase!

idrysdale:
On it!

idrysdale:
Done it! Ready for eyes 👀

database/database.go
```diff
@@ -163,6 +163,24 @@ func (db *Database) DeleteRequestToJoinTeam(teamUUID uuid.UUID, fingerprint fpr.
 	return db.saveToFile(*message)
 }
 
+// GetRequestsToJoinTeam returns a slice of requests to join the team matching the UUID.
+func (db *Database) GetRequestsToJoinTeam(teamUUID uuid.UUID) (
```
paulfurley: What's the logic behind it returning *multiple* when we actually want to de-duplicate and only get the *earliest* ?

paulfurley: Looking at the other PR it seems we *actually* want a function that just queries if a request already exists, rather than actually getting a list of them, right?

idrysdale: You're right, see https://github.com/fluidkeys/fluidkeys/pull/405

PR #398, opened 2019-03-18
Ensure you can't request to join a team you've already asked to join!

idrysdale:
Check the database and exit during `team join` if there are any existing requests to join the same team.

(Requires https://github.com/fluidkeys/fluidkeys/pull/397)

### DIFF
--- a/fk/teamjoin.go
+++ b/fk/teamjoin.go
@@ -31,6 +31,10 @@ import (
 )
 
 func teamJoin(teamUUID uuid.UUID) exitCode {
+	if exitCode := ensureNoExistingRequests(teamUUID); exitCode != 0 {
+		return exitCode
+	}
+
 	teamName, err := client.GetTeamName(teamUUID)
 	if err != nil {
 		out.Print(ui.FormatFailure("Couldn't request to join team", nil, err))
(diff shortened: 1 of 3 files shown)


fk/teamjoin.go
```diff
@@ -69,6 +75,27 @@ func teamJoin(teamUUID uuid.UUID) exitCode {
...
+	existingRequests, err := db.GetRequestsToJoinTeam(teamUUID)
+	if err != nil {
+		out.Print(ui.FormatFailure("Failed to check for existing requests", nil, err))
+		return 1
+	}
+	if len(existingRequests) > 0 {
+		out.Print(ui.FormatWarning(
+			"You've already requested to join "+existingRequests[0].TeamName,
+			[]string{
+				"You requested to join " + existingRequests[0].TeamName + " " +
```
paulfurley: be great if you could use the helper function I've made for this, I'll put the commit as a separate pull request

idrysdale: Wilco!

idrysdale: Done!

fk/teamjoin.go
```diff
@@ -31,6 +31,10 @@ import (
 )
 
 func teamJoin(teamUUID uuid.UUID) exitCode {
+	if exitCode := ensureNoExistingRequests(teamUUID); exitCode != 0 {
```
paulfurley: Doesn't this want to be *after* we've worked out which key they're using? A request to join a key is unique on `(team UUID, fingerprint)`, not on `teamUUID`

idrysdale: :) good thinking

PR #401, opened 2019-03-18
add team.VerifyRoster(roster, signature, adminKeys)

paulfurley:


team/team_test.go
```diff
@@ -84,9 +84,11 @@ name = "Kiffix"
 		})
 
 		t.Run("sets a valid signature", func(t *testing.T) {
-			verifyRosterSignature(t,
```
idrysdale: suspect you could also drop `func verifyRosterSignature()` now?

paulfurley: Ah great spot, I missed out that commit from this PR!

paulfurley: Actually I just forgot to delete it!

paulfurley: Done

paulfurley:
rebased

PR #402, opened 2019-03-18
add client.DeleteRequestToJoinTeam

idrysdale:
I'll use this later so that when admins are reviewing requests, they delete them.

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -760,6 +760,50 @@ func TestListRequestsToJoinTeam(t *testing.T) {
 	})
 }
 
+func TestDeleteRequestToJoinTeam(t *testing.T) {
+	t.Run("parses the name from a good response", func(t *testing.T) {
+		client, mux, _, teardown := setup()
+		defer teardown()
+
+		teamUUID := uuid.Must(uuid.NewV4())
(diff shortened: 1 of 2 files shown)


api/api_test.go
```diff
@@ -760,6 +760,52 @@ func TestListRequestsToJoinTeam(t *testing.T) {
...
+	t.Run("passes us error codes", func(t *testing.T) {
+		client, mux, _, teardown := setup()
+		defer teardown()
+
+		teamUUID := uuid.Must(uuid.NewV4())
+		unknownRequestUUID := uuid.Must(uuid.NewV4())
+
+		mockResponseHandler := func(w http.ResponseWriter, r *http.Request) {
+			assertClientSentVerb(t, "DELETE", r.Method)
+			w.Header().Add("Content-Type", "application/json")
```
paulfurley: The API doesn't send `application/json` for this endpoint (it doesn't send any body)

idrysdale: Ah, yes, good spot!

paulfurley: You can self-merge when this is changed (in 2 places: the other test, too)

PR #404, opened 2019-03-18
fk team sync: get & verify new roster & keys

paulfurley:
this handles the happy path and *most* other paths for joining a new team
(based on requests to join a team)

things to note:

* new rosters are cryptographically validated. this requires downloading
 the admin keys from the API by fingerprint. the behaviour for
 validating an *update* will be different

* requests time out after 7 days. this is currently implemented in
 `fk team sync` but I think I'll move it to database in a subsequent PR
 (along with de-duplicating requests)

* requests are deleted from our local db if they succeed

* content should match our Google doc:
https://docs.google.com/document/d/1FHQXHP7X8Uvtvn72OqZNVYyFMocrJ2MM1zWtYTM4d0c/edit#heading=h.htvwu2p78dq

### DIFF
--- a/fk/teamsync.go
+++ b/fk/teamsync.go
@@ -20,6 +20,7 @@ package fk
 import (
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/fluidkeys/fluidkeys/api"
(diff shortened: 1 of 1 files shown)


paulfurley:
rebased: those `DELETEME` commits now deleted

PR #406, opened 2019-03-18
team join: you've already requested to join that team

idrysdale:
This should be fine, now that https://github.com/fluidkeys/fluidkeys/pull/411 is in.

### DIFF
--- a/fk/teamjoin.go
+++ b/fk/teamjoin.go
@@ -44,6 +44,10 @@ func teamJoin(teamUUID uuid.UUID) exitCode {
 		return code
 	}
 
+	if exitCode := ensureNoExistingRequests(teamUUID, pgpKey.Fingerprint()); exitCode != 0 {
+		return exitCode
+	}
+
 	email, err := pgpKey.Email()
 	if err != nil {
 		out.Print(ui.FormatFailure("Error getting email for key", nil, err))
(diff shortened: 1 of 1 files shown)


fk/teamjoin.go
```diff
@@ -69,6 +73,26 @@ func teamJoin(teamUUID uuid.UUID) exitCode {
 
 }
 
+func ensureNoExistingRequests(teamUUID uuid.UUID, fingerprint fpr.Fingerprint) exitCode {
+	existingRequest, err := db.GetExistingRequestToJoinTeam(teamUUID, fingerprint)
+	if err != nil {
```
paulfurley: this catches `ErrRequestNotFound` and therefore prevents you getting to line 82:

image

idrysdale: Good spot!

Do you think having this `ErrRequestNotFound` on the query is correct?

Perhaps we should just return no error, but nil request?

paulfurley: `nil` request with no error works for me

idrysdale: 👍

idrysdale: Because of this change https://github.com/fluidkeys/fluidkeys/pull/411 this error won't raise on no errors now.

PR #408, opened 2019-03-18
fix Golint errors

paulfurley:
...introduced in 142afe9c

note that error strings should begin with lowecase


idrysdale:
Thank you!

PR #409, opened 2019-03-18
de-duplicate in RecordRequestToJoinTeam, GetRequestsToJoinTeams

paulfurley:
do a bit of test tidy-up:
* delete zero-value initializations: they have no effect, and just distract from the tests
* factor out `now` and `later`


idrysdale:
👍

PR #416, opened 2019-03-19
Make `IsInTeam` return a pointer to the team

idrysdale:
This means that that calling function can do things like:

```
isInTeam, team, err := user.IsInTeam(uuid)
...
if isInTeam {
   fmt.Printf("You're already in %s", team.Name)
}
```

### DIFF
--- a/user/user_test.go
+++ b/user/user_test.go
@@ -79,13 +79,33 @@ func TestMembershipFunctions(t *testing.T) {
 	})
 
 	t.Run("InTeam", func(t *testing.T) {
-		got, err := user.IsInTeam(team1.UUID)
-		assert.NoError(t, err)
-		assert.Equal(t, true, got)
+		t.Run("user is in team", func(t *testing.T) {
+			gotIsInTeam, gotTeam, err := user.IsInTeam(team1.UUID)
+			assert.NoError(t, err)
 
-		got, err = user.IsInTeam(team2.UUID)
-		assert.NoError(t, err)
-		assert.Equal(t, false, got)
+			t.Run("returns true", func(t *testing.T) {
+				assert.Equal(t, true, gotIsInTeam)
+			})
+
+			t.Run("returns a pointer to the team", func(t *testing.T) {
+				assert.Equal(t, &team1, gotTeam)
+			})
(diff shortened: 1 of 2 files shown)


user/user_test.go
```diff
@@ -79,13 +79,33 @@ func TestMembershipFunctions(t *testing.T) {
...
-		assert.Equal(t, false, got)
+		t.Run("user is in team", func(t *testing.T) {
+			gotIsInTeam, gotTeam, err := user.IsInTeam(team1.UUID)
+			assert.NoError(t, err)
+
+			t.Run("returns true", func(t *testing.T) {
+				assert.Equal(t, true, gotIsInTeam)
+			})
+
+			t.Run("returns a pointer to the team", func(t *testing.T) {
```
paulfurley: :+1:

user/user_test.go
```diff
@@ -79,13 +79,33 @@ func TestMembershipFunctions(t *testing.T) {
...
+
+		t.Run("user is not in team", func(t *testing.T) {
+			gotIsInTeam, gotTeam, err := user.IsInTeam(team2.UUID)
+			assert.NoError(t, err)
+
+			t.Run("returns true", func(t *testing.T) {
+				assert.Equal(t, false, gotIsInTeam)
+			})
+
+			t.Run("returns nil", func(t *testing.T) {
```
paulfurley: "returns nil" would be clearer if it were "return value of theTeam is nil"

user/user_test.go
```diff
@@ -79,13 +79,33 @@ func TestMembershipFunctions(t *testing.T) {
...
+			t.Run("returns a pointer to the team", func(t *testing.T) {
+				assert.Equal(t, &team1, gotTeam)
+			})
+		})
+
+		t.Run("user is not in team", func(t *testing.T) {
+			gotIsInTeam, gotTeam, err := user.IsInTeam(team2.UUID)
+			assert.NoError(t, err)
+
+			t.Run("returns true", func(t *testing.T) {
```
paulfurley: "returns true" would be clear if it were "return value of isInTeam is true"

PR #417, opened 2019-03-19
team join: you're already in that team

idrysdale:


fk/teamjoin.go
```diff
@@ -31,6 +31,22 @@ import (
...
 
 func teamJoin(teamUUID uuid.UUID) exitCode {
+	isInTeam, existingTeam, err := user.IsInTeam(teamUUID)
+	if err != nil {
+		out.Print(ui.FormatFailure("Couldn't check if user already is in team", nil, err))
+		return 1
+	}
+	if isInTeam {
+		out.Print(ui.FormatWarning(
+			"You're already in the team"+existingTeam.Name,
```
paulfurley: missing space here 
image

fk/teamjoin.go
```diff
@@ -31,6 +31,22 @@ import (
...
+			[]string{
+				"You can't request to join a team you're already part of.",
```
paulfurley: I'd suggest to drop this extra line as it's obvious.

I'd also suggest making it a *positive* message. They've tried to join the team, and hooray, they're in the team already. It's a celebrate not a warning.

Perhaps better to signpost them to another command? What might they have been trying to do when they joined twice?

PR #418, opened 2019-03-19
team authorize: try to delete all requests at end

idrysdale:


fk/teamauthorize.go
```diff
@@ -92,19 +100,9 @@ func teamAuthorize() exitCode {
 	}
 }
 
-func reviewRequests(myTeam team.Team, adminKey pgpkey.PgpKey) (
-	approvedRequests []team.RequestToJoinTeam, code exitCode) {
+func reviewRequests(requests []team.RequestToJoinTeam, myTeam team.Team) (
```
paulfurley: it's nice that this can no longer error

paulfurley:
I found it a little surprising that if you answer "no" that's irreversible: the request is deleted without any further prompt.

paulfurley:
But it's probably cool!

PR #420, opened 2019-03-19
Block `team create` and `join` if already in team

idrysdale:
I _think_ these are the only two ways to create a membership from
the command line. I wonder if I need to also check elsewhere in
the application? For example, what would happen if the user
hacked their `fluidkeys/teams` directory, and duplicated a team
subdirectory?

### DIFF
--- a/fk/teamcreate.go
+++ b/fk/teamcreate.go
@@ -33,6 +33,10 @@ import (
 )
 
 func teamCreate() exitCode {
+	if code := ensureUserIsntInATeam("Couldn't create team"); code > 0 {
+		return code
+	}
+
 	out.Print("\n")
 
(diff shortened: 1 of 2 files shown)


paulfurley:
I'll rebase this now and review it.

fk/teamcreate.go
```diff
@@ -33,6 +33,10 @@ import (
 )
 
 func teamCreate() exitCode {
+	if code := ensureUserIsntInATeam("Couldn't create team"); code > 0 {
```
paulfurley: everywhere else we use `exitCode != 0` to test it, why use `exitCode > 0` now?

are you *sure* you mean "is a positive number" rather than "isn't zero"? (I have no idea if negative return codes are possible, but it wouldn't surprise me)

PR #422, opened 2019-03-19
team fetch: fetch, verify and replace team roster

paulfurley:


fk/teamfetch.go
```diff
@@ -256,16 +256,31 @@ func getAndImportKeyToGpg(fingerprint fp.Fingerprint) error {
...
-			return nil, fmt.Errorf("failed to get admin key %s: %v", p.Fingerprint, err)
+			return nil, err
 		}
-
 		adminKeys = append(adminKeys, key)
 	}
 	return adminKeys, nil
 }
 
+func discoverPublicKey(fingerprint fp.Fingerprint) (key *pgpkey.PgpKey, err error) {
```
idrysdale: Niiiiiice 👌

fk/teamfetch.go
```diff
@@ -55,13 +55,14 @@ func teamFetch() exitCode {
 			continue
 		}
 
-		if err := fetchAndUpdateRoster(membership.Team, unlockedKey); err != nil {
+		updatedTeam, err := fetchAndUpdateRoster(membership.Team, unlockedKey)
+		if err != nil {
 			out.Print(ui.FormatWarning("Failed to check team for updates", []string{}, err))
 			sawError = true
 			continue
```
paulfurley: oof, just because we failed to update the team doesn't mean we shouldn't carry on and fetch our team's keys

PR #426, opened 2019-03-20
add `fk sync` command

paulfurley:
this just wraps 2 commands:

* `fk key maintain automatic`
* `fk team fetch`

### DIFF
--- a/fk/sync.go
+++ b/fk/sync.go
@@ -1,0 +1,45 @@
+// Copyright 2019 Paul Furley and Ian Drysdale
+//
+// This file is part of Fluidkeys Client which makes it simple to use OpenPGP.
+//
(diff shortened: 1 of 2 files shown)


fk/sync.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"log"
+
+	"github.com/docopt/docopt-go"
+	"github.com/fluidkeys/fluidkeys/colour"
+	"github.com/fluidkeys/fluidkeys/out"
+)
+
+func syncSubcommand(args docopt.Opts) exitCode {
+	dryRun, err := args.Bool("--dry-run")
```
idrysdale: I think this can be removed!

paulfurley: done!

PR #434, opened 2019-03-21
Add db.RecordLast and GetLast

idrysdale:
These functions can be used to record when certain actions get performed on certain items.

This PR supports:

* fingerprints and keys (which are treated the same)
* teams

### DIFF
--- a/database/database.go
+++ b/database/database.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	fpr "github.com/fluidkeys/fluidkeys/fingerprint"
+	"github.com/fluidkeys/fluidkeys/pgpkey"
 	"github.com/fluidkeys/fluidkeys/team"
 	"github.com/gofrs/uuid"
 )
@@ -41,6 +42,7 @@ type Database struct {
 type Message struct {
 	KeysImportedIntoGnuPG []KeyImportedIntoGnuPGMessage
 	RequestsToJoinTeams   []RequestToJoinTeamMessage
+	EventTimes            map[string]time.Time
 }
 
 // KeyImportedIntoGnuPGMessage represents a key the user has imported into GnuPG from Fluidkeys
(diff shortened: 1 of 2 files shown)


database/database.go
```diff
@@ -105,6 +107,57 @@ func (db *Database) RecordRequestToJoinTeam(
...
+		return message.LastUpdated[verb+":"+keyItem+":"+i.Uri()], nil
+	case fpr.Fingerprint:
+		return message.LastUpdated[verb+":"+keyItem+":"+i.Uri()], nil
+	case *team.Team:
+		return message.LastUpdated[verb+":"+teamItem+":"+i.UUID.String()], nil
+	case team.Team:
+		return message.LastUpdated[verb+":"+teamItem+":"+i.UUID.String()], nil
+	}
+
+	return time.Time{}, fmt.Errorf("no record of when %v was last '%s'", item, verb)
```
paulfurley: it would be helpful for my use case to tell the difference between an error and "everything OK, but key not found".

suggest a custom error type `ErrNotFound`

database/database.go
```diff
@@ -41,6 +42,7 @@ type Database struct {
 type Message struct {
 	KeysImportedIntoGnuPG []KeyImportedIntoGnuPGMessage
 	RequestsToJoinTeams   []RequestToJoinTeamMessage
+	LastUpdated           map[string]time.Time
```
paulfurley: The name should probably now be e.g. `EventTimes` or similar

paulfurley:
looking forward to using this :D

database/database_test.go
```diff
@@ -344,6 +347,113 @@ func TestGetExistingRequestToJoinTeam(t *testing.T) {
...
+		t.Run("record updates the previously time", func(t *testing.T) {
+			err := database.RecordLast("fetch", fingerprint, later)
+			assert.NoError(t, err)
+
+			got, err := database.GetLast("fetch", fingerprint)
+			assert.NoError(t, err)
+			assert.Equal(t, later, got)
+		})
+
+		t.Run("a second value records it's own time", func(t *testing.T) {
```
paulfurley: what does it mean, "it's own time" ?

database/database_test.go
```diff
@@ -344,6 +347,113 @@ func TestGetExistingRequestToJoinTeam(t *testing.T) {
...
+			assert.Equal(t, now, got)
+		})
+
+		t.Run("pointers and values are handled the same", func(t *testing.T) {
+			got, err := database.GetLast("fetch", &fingerprint)
+			assert.NoError(t, err)
+			assert.Equal(t, now, got)
+		})
+
+		t.Run("record updates the previously time", func(t *testing.T) {
```
paulfurley: does "updates the previously time" mean "replaces the existing time" ?

database/database_test.go
```diff
@@ -344,6 +347,113 @@ func TestGetExistingRequestToJoinTeam(t *testing.T) {
...
+			got, err := database.GetLast("fetch", fingerprint)
+			assert.NoError(t, err)
+			assert.Equal(t, later, got)
+
+			got, err = database.GetLast("fetch", fingerprint2)
+			assert.NoError(t, err)
+			assert.Equal(t, now, got)
+		})
+
+		t.Run("keys are handled the same", func(t *testing.T) {
```
paulfurley: the same as what?

paulfurley:
Yay!

merging with 2 suggestions:
* make the test names less confusing
* add a test case to cover when there is no record (it should return zero time and error=nil, as we discussed)

PR #441, opened 2019-03-22
`fk status`: Print keys and team members for memberships

idrysdale:
This PR handles all of a user's memberships, or rather all keys they are members of a team with.

If prints something that looks like this:
Screenshot 2019-03-21 at 19 04 28

### DIFF
--- a/fk/status.go
+++ b/fk/status.go
@@ -50,6 +50,13 @@ func statusSubcommand(args docopt.Opts) exitCode {
 		for _, membership := range groupedMembership.Memberships {
 			key, err := loadPgpKey(membership.Me.Fingerprint)
 			if err != nil {
+				out.Print(ui.FormatFailure(
+					"Failed to load key associated with team "+membership.Team.Name,
+					[]string{
+						"Tried to load key " + membership.Me.Fingerprint.Hex(),
+					},
+					err,
+				))
 				return 1
 			}
 
--- a/fk/status.go
+++ b/fk/status.go
@@ -18,7 +18,6 @@
 package fk
 
 import (
-	"log"
 	"time"
 
 	docopt "github.com/docopt/docopt-go"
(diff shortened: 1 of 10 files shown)


fk/status.go
```diff
@@ -0,0 +1,107 @@
...
+		printHeader(groupedMembership.Team.Name)
+
+		adminOfTeam := false
+
+		teamKeysWithWarnings := []table.KeyWithWarnings{}
+
+		for _, membership := range groupedMembership.Memberships {
+			key, err := loadPgpKey(membership.Me.Fingerprint)
+			if err != nil {
+				return 1
```
paulfurley: error swallowed: should be `out.Print(ui.FormatFailure(..))`

fk/status.go
```diff
@@ -18,10 +18,60 @@
...
+	"github.com/fluidkeys/fluidkeys/status"
+	"github.com/fluidkeys/fluidkeys/table"
 )
 
 func statusSubcommand(args docopt.Opts) exitCode {
+	out.Print("\n")
+
+	groupedMemberships, err := user.GroupedMemberships()
+	if err != nil {
+		log.Panic(err)
```
paulfurley: eek

paulfurley: i closed this card once! :P

paulfurley: i've pushed a commit to fix this

fk/status.go
```diff
@@ -85,6 +85,8 @@ func statusSubcommand(args docopt.Opts) exitCode {
 				colour.Cmd("fk team join " + groupedMembership.Team.UUID.String()),
 			}))
 		}
+
+		out.Print("To fetch and store team members keys run " + colour.Cmd("fk team fetch") + "\n\n")
```
paulfurley: How about "The team's keys are updated automatically. To check for updates now, run fk team fetch"

idrysdale: Sounds good to me!

paulfurley:
I think `Never` would look better as either `never` or `-`

and `Admin` would look better as `admin`

idrysdale:
Contenious, but you can have it 😉

paulfurley:
AWESOME

PR #444, opened 2019-03-22
`fk status`: Print requests to join teams

idrysdale:
Looks like this:
Screenshot 2019-03-22 at 16 00 14

### DIFF
--- a/fk/status.go
+++ b/fk/status.go
@@ -36,6 +36,9 @@ func statusSubcommand(args docopt.Opts) exitCode {
 		return code
 	}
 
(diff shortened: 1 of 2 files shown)


paulfurley:
image

It's a bit misleading when I'm in the following state:
* I requested to join a team
* My request has been approved
* I run `fk status` ...

so `fk status` tells me my request hasn't been approved, but it's no longer true. 

perhaps the simple fix is if it tells me something like

```
Fluidkeys will periodically check if your request has been approved, or you can check now by running
fk team fetch
```

PR #446, opened 2019-03-22
`fk status`: Print orphaned keys

idrysdale:
Rebase after https://github.com/fluidkeys/fluidkeys/pull/444

Looks like this:
Screenshot 2019-03-22 at 17 47 46

### DIFF
--- a/fk/status.go
+++ b/fk/status.go
@@ -39,6 +39,10 @@ func statusSubcommand(args docopt.Opts) exitCode {
 	if code := printRequests(); code != 0 {
 		return code
 	}
+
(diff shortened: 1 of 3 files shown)


paulfurley:
rebasing

paulfurley:
image

^^ It's a bit overwhelming getting the "You're not in a team" message 3 times.

A header in between would help. Or only showing it once?

I'd actually expected these orphaned keys to show up in a table basically matching `fk key list`

idrysdale:
I've updated it, so it should now look more like this:

Screenshot 2019-03-25 at 11 28 55

paulfurley:
testing now!

PR #449, opened 2019-03-22
team fetch: only update roster & keys every 24hrs

paulfurley:
if we're running in *unattended* mode (e.g. from cron via `fk sync`)

(if we're running `fk team fetch` directly then *always* download updates)

### DIFF
--- a/fk/teamfetch.go
+++ b/fk/teamfetch.go
@@ -61,6 +61,8 @@ func teamFetch(unattended bool) exitCode {
 }
 
 func doUpdateTeam(myTeam *team.Team, me *team.Person, unattended bool) (err error) {
+	alwaysDownload := !unattended
+
 	printHeader(myTeam.Name)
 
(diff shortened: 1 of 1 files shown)


fk/teamfetch.go
```diff
@@ -101,19 +103,34 @@ func formatYouRequestedToJoin(request team.RequestToJoinTeam) string {
 		humanize.RoughDuration(time.Now().Sub(request.RequestedAt)) + " ago."
 }
 
-func fetchAndUpdateRoster(t team.Team, unlockedKey *pgpkey.PgpKey) (
+// fetchAndUpdateRoster fetches any update to the team roster and saves it back to disk.
+// is alwaysDownload is false, only check the roster if we last checked it more than 24 hours ago
```
idrysdale: is/if typo

paulfurley: thanks!

PR #450, opened 2019-03-25
factor out formatFailedToLoadKey, add help text

paulfurley:
help the user recover from "Failed to load key" message by signposting
their db.json file


paulfurley:
Rebased and ready.

PR #458, opened 2019-03-25
Implement redesign of `team apply`:

idrysdale:
* dont print checkbox about `Request to join...`
* add info box telling user to reply
* print message to apply with
* prompt to copy this message to clipboard

### DIFF
--- a/fk/teamapply.go
+++ b/fk/teamapply.go
@@ -58,17 +58,13 @@ func teamApply(teamUUID uuid.UUID) exitCode {
 		return 1
 	}
 
-	printHeader("Requesting to join team")
-
-	action := "Request to join " + teamName
-	ui.PrintCheckboxPending("action")
+	printHeader("Apply to join team")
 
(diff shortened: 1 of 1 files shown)


fk/teamapply.go
```diff
@@ -69,8 +69,23 @@ func teamApply(teamUUID uuid.UUID) exitCode {
 		"This information allows them to verify your request.",
 	}))
 
-	out.Print("Your team admin will need to authorize your request for Fluidkeys to\n" +
-		"start working.\n\n")
+	out.Print(formatFileDivider("Please authorize me to join Kiffix", 80) + "\n")
+	requestMessage := `I've requested to join ` + teamName + ` on Fluidkeys.
```
paulfurley: i think we probably shouldn't use the raw string operator in the middle of code like this as it makes it quite hard to see the structure of indents. Nothing wrong with "...\n" I reckon

paulfurley: I've added a commit for this.

paulfurley:
I added a commit to use the friendly fingerprint format (not `.Hex()`) to match `team authorize`

PR #462, opened 2019-03-25
team authorize: if "no", prompt to reject request

paulfurley:
* make instructions refer to verification details and be clearer what we mean by "check"
* "Authorize keys" -> "Authorize requests to join ...


idrysdale:
This looks great! The only bit I'm not sure of is the word 'block'. What made you pick that?

(Alternatives could be reject, deny, dismiss...)

PR #467, opened 2019-03-25
If no keys, `status` should prompt to run `setup`

idrysdale:


fk/status.go
```diff
@@ -70,6 +70,13 @@ func statusSubcommand(args docopt.Opts) exitCode {
 	}
 	allKeysWithWarnings = append(allKeysWithWarnings, orphanedKeysWithWarnings...)
 
+	if len(allKeysWithWarnings) == 0 {
+		out.Print(ui.FormatInfo("Get started with Fluidkeys", []string{
+			"You should run " + colour.Cmd("fk setup"),
```
paulfurley: I think this is a missed opportunity (and it's kind of abrupt)

Suggest:
```
[i] Get started with Fluidkeys

If your team are already using Fluidkeys, ask your team's admin for the
team ID. Join the team by running fk apply <team id>

Create a team with a brand new PGP key by running fk team create

To use Fluidkeys on your own, get started by running fk setup
```

PR #474, opened 2019-04-30
extend keys annually & extend (don't rotate) subkeys

paulfurley:
after extensive testing, we've realised the world isn't ready for fast-expiring PGP keys!

this PR introduces the following changes:

1. keys last ~1 year
   previously, keys expired 30 after 1st of next month
   now, keys expire in 1 year *rounded* to the next quarter (1st feb, may, august or november)
   This means there will only be 4 dates in the year when expiries occur, allowing teams to
   help each other out 

2. extend (don't rotate) subkeys
   creating new subkeys can cause trouble with multiple devices, and using a single password
   for primary & subkeys is a false sense of security.
   instead, just extend subkeys the same as the primary key

3. start trying to extend 60 days before the expiry date (previously it was 30)
   so if your expiry date is 1st May, on 2nd March (60 days before), fk will start trying to
   extend your key. This gives us *plenty* of time to notify people if that isn't working
   (e.g. at -45 days, which is still well before 30 day warnings from e.g. Enigmail)

* implements this card: https://trello.com/c/yA2hEInD
* fixes this bug (enigmail warning): https://trello.com/c/hRS7LE7D

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -155,11 +155,12 @@ const (
 )
 
 // NextExpiryTime returns the expiry time in UTC, according to the policy:
-//     "30 days after the 1st of the next month"
-// for example, if today is 15th September, nextExpiryTime would return
-// 1st October + 30 days
+//     "1 year from now, rounded forward to the 1st of the next Feb, May, Aug or Nov
+// for example, if today is 15th September 2018, nextExpiryTime would return
+// 1st November 2019
 func NextExpiryTime(now time.Time) time.Time {
-	return firstOfNextMonth(now).Add(thirtyDays).In(time.UTC)
+	oneYearFromNow := now.In(time.UTC).Add(oneYear)
+	return followingQuarter(oneYearFromNow)
 }
 
 // NextRotation returns 30 days before the earliest expiry time on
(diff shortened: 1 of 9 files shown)


policy/policy.go
```diff
@@ -207,4 +228,5 @@ const (
 	tenDays       time.Duration = time.Duration(time.Hour * 24 * 10)
 	thirtyDays    time.Duration = time.Duration(time.Hour * 24 * 30)
 	fortyFiveDays time.Duration = time.Duration(time.Hour * 24 * 45)
+	oneYear       time.Duration = time.Duration(time.Hour * 24 * 365)
```
idrysdale: Is there any reason to be concerned about leap years having 366 days?

paulfurley: Yes, but check out the tests!

tldr: the algorithm for finding the following quarter accounts for this.

status/keyactions.go
```diff
@@ -72,23 +72,24 @@ func (a CreateNewEncryptionSubkey) SortOrder() int {
...
-func (a ExpireSubkey) Enact(key *pgpkey.PgpKey, now time.Time, password *string) error {
-	return key.ExpireSubkey(a.SubkeyId, now)
+func (a ModifySubkeyExpiry) Enact(key *pgpkey.PgpKey, now time.Time, password *string) error {
+	return key.UpdateSubkeyValidUntil(a.subkeyId, a.validUntil, now)
 }
-func (a ExpireSubkey) String() string {
-	return fmt.Sprintf("Expire the encryption subkey now (0x%X)", a.SubkeyId)
+
+func (a ModifySubkeyExpiry) String() string {
+	return fmt.Sprintf("Extend encryption subkey expiry to %s", a.validUntil.Format("2 Jan 06"))
```
paulfurley: @idrysdale what did this one do to you? :stuck_out_tongue:

PR #477, opened 2019-05-01
team fetch: only certify keys once

paulfurley:
the end goal here is that *normally*, `fk sync` doesn't require an unlocked
key.

that's good because:

1. we shouldn't unlock a key if it's not strictly necessary
2. we can add error reporting to the unlock key routine, and we'll only
  see errors when they actually matter (e.g. a new team member has
  joined the team, and their key needs certifying)


paulfurley:
I've realised this isn't right because a `recordThatKeyWasCertified` should *also* include the email address. We certifiy *UIDSs* not *keys*.

PR #478, opened 2019-05-01
rename api / client / apiclient

paulfurley:
1. rename `api` package to `apiclient`: it's not an api, it's a client of an API!

2. rename global `client` -> `api` in `fk/main.go`
   this allows the more natural feeling `api.GetPublicKey(..)`

3. rename `apiclient.NewClient` -> `apiclient.New`


idrysdale:
Not contentious! Sensible!

PR #480, opened 2019-05-01
Introduce launchd

idrysdale:
I made a schedulerInterface and seperated out cron and launchd.

We check for `launchd` in the init() of the scheduler package
to figure out whether to pass through to launchd or cron.

### DIFF
--- a/fk/main.go
+++ b/fk/main.go
@@ -97,7 +97,7 @@ Options:
 	log.Print("$ " + strings.Join(os.Args, " "))
 	args, _ := docopt.ParseDoc(usage)
 
-	ensureCrontabStateMatchesConfig()
+	ensureSchedulerStateMatchesConfig()
 
 	cronOutput, err := args.Bool("--cron-output")
 	if err != nil {
@@ -144,57 +144,81 @@ Options:
 	return code
 }
 
-func ensureCrontabStateMatchesConfig() {
+func ensureSchedulerStateMatchesConfig() {
 	shouldEnable, err := shouldEnableScheduler()
 	if err != nil {
 		log.Panic(err)
 	}
 
 	if shouldEnable {
-		crontabWasAdded, err := scheduler.Enable(nil)
+		schedulerWasEnabled, err := scheduler.Enable()
 		if err != nil {
-			out.Print(ui.FormatFailure(
-				"Failed to schedule automatic key maintenance and rotation", []string{
-					"Fluidkeys manages your key by running itself periodically with cron.",
-					"Something prevented Fluidkeys from adding itself to your crontab."},
-				err,
-			))
+			switch err.(type) {
+			case *scheduler.ErrModifyingCrontab:
+				out.Print(ui.FormatFailure(
+					"Failed to schedule automatic key maintenance and rotation", []string{
+						"Fluidkeys manages your key by running itself periodically with cron.",
+						"Something prevented Fluidkeys from adding itself to your crontab."},
+					err,
+				))
 
-			out.Print("To fix this, run " + colour.Cmd("crontab -e") + " and add these lines:\n\n")
-			out.Print(formatFileDivider("crontab", 80))
-			out.Print("\n" + scheduler.CronLines)
-			out.Print(formatFileDivider("", 80))
-			out.Print("\n\n")
+				out.Print("To fix this, run " + colour.Cmd("crontab -e") + " and add these lines:\n\n")
+				out.Print(formatFileDivider("crontab", 80))
+				out.Print("\n" + scheduler.CronLines)
+				out.Print(formatFileDivider("", 80))
+				out.Print("\n\n")
+
+			default:
+				out.Print(ui.FormatFailure(
+					"Failed to schedule automatic key maintenance and rotation", []string{
+						"Fluidkeys manages your key by running itself periodically with " +
+							scheduler.Name(),
+						"Something prevented this happening."},
+					err,
+				))
+			}
 
 			// don't carry on: they need to fix this problem first.
 			// if automatic key maintenance isn't working, Fluidkeys can't work
 			os.Exit(1)
 		}
 
-		if crontabWasAdded {
-			printInfo(fmt.Sprintf("Added Fluidkeys to crontab.  Edit %s to remove.",
-				Config.GetFilename()))
+		if schedulerWasEnabled {
+			printInfo(fmt.Sprintf("Added Fluidkeys to %s.  Edit %s to remove.",
+				scheduler.Name(), Config.GetFilename()))
 		}
 	} else {
-		crontabWasRemoved, err := scheduler.Disable(nil)
+		schedulerWasDisabled, err := scheduler.Disable()
 		if err != nil {
-			out.Print(ui.FormatWarning(
-				"Failed to remove Fluidkeys from crontab", []string{
-					"Fluidkeys tried to remove itself from crontab but something prevented it.",
-				},
-				err,
-			))
+			switch err.(type) {
+			case *scheduler.ErrModifyingCrontab:
+				out.Print(ui.FormatWarning(
+					"Failed to remove Fluidkeys from crontab", []string{
+						"Fluidkeys tried to remove itself from crontab but something prevented it.",
+					},
+					err,
+				))
 
-			out.Print("To fix this, run " + colour.Cmd("crontab -e") + " and remove these lines:\n\n")
-			out.Print(formatFileDivider("crontab", 80))
-			out.Print("\n" + scheduler.CronLines)
-			out.Print(formatFileDivider("", 80))
-			out.Print("\n\n")
+				out.Print("To fix this, run " + colour.Cmd("crontab -e") + " and remove these lines:\n\n")
+				out.Print(formatFileDivider("crontab", 80))
+				out.Print("\n" + scheduler.CronLines)
+				out.Print(formatFileDivider("", 80))
+				out.Print("\n\n")
+
+			default:
+				out.Print(ui.FormatFailure(
+					"Failed to disable automatic key maintenance and rotation", []string{
+						"Fluidkeys manages your key by running itself periodically with " +
+							scheduler.Name(),
+						"Something prevented this from being disabled."},
+					err,
+				))
+			}
 		}
 
-		if crontabWasRemoved {
-			printInfo(fmt.Sprintf("Removed Fluidkeys from crontab.  Edit %s to add again.",
-				Config.GetFilename()))
+		if schedulerWasDisabled {
+			printInfo(fmt.Sprintf("Removed Fluidkeys from %s.  Edit %s to add again.",
+				scheduler.Name(), Config.GetFilename()))
 		}
 	}
 }
--- a/scheduler/cron.go
+++ b/scheduler/cron.go
@@ -1,0 +1,190 @@
+package scheduler
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os/exec"
+	"strings"
+)
+
+type cron struct{}
+
+// Enable writes Fluidkeys' cron lines into crontab
+func (c *cron) Enable() (crontabWasAdded bool, err error) {
+	crontab := &systemCrontab{}
+	return c.enable(crontab)
+}
+
+// Disable parses the crontab (output of `crontab -l`) and removes Fluidkeys'
+// cron lines if present.
+// If the remaining crontab is empty, the crontab is removed with `crontab -r`
+func (c *cron) Disable() (cronLinesWereRemoved bool, err error) {
+	crontab := &systemCrontab{}
+	return c.disable(crontab)
+}
+
+func (c *cron) Name() string {
+	return "crontab"
+}
+
+func (c *cron) enable(crontab runCrontabInterface) (crontabWasAdded bool, err error) {
+	currentCrontab, err := crontab.get()
+	if err != nil {
+		return false, fmt.Errorf("error getting crontab: %v", err)
+	}
+
+	if !hasFluidkeysCronLines(currentCrontab) {
+		newCrontab := addCrontabLinesWithoutRepeating(currentCrontab)
+		err = crontab.set(newCrontab)
+		if err != nil {
+			return false, ErrModifyingCrontab{origError: err}
+		}
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func (c *cron) disable(crontab runCrontabInterface) (cronLinesWereRemoved bool, err error) {
+	currentCrontab, err := crontab.get()
+	if err != nil {
+		return false, fmt.Errorf("error getting crontab: %v", err)
+	}
+
+	if hasFluidkeysCronLines(currentCrontab) {
+		newCrontab := removeCrontabLines(currentCrontab)
+		err = crontab.set(newCrontab)
+
+		if err != nil {
+			return false, ErrModifyingCrontab{origError: err}
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
+func hasFluidkeysCronLines(crontab string) bool {
+	return strings.Contains(crontab, strings.TrimSuffix(CronLines, "\n"))
+}
+
+type systemCrontab struct{}
+
+func (s *systemCrontab) get() (string, error) {
+	output, err := s.runCrontab("-l")
+
+	if s.isNoCrontabError(output, err) {
+		return "", nil
+	}
+
+	return output, err
+}
+
+func (s *systemCrontab) set(newCrontab string) error {
+	f, err := ioutil.TempFile("", "")
+	if err != nil {
+		return fmt.Errorf("error opening temp file: %v", err)
+	}
+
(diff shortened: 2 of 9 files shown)


fk/main.go
```diff
@@ -144,57 +144,63 @@ Options:
...
-func ensureCrontabStateMatchesConfig() {
+func ensureSchedulerStateMatchesConfig() {
 	shouldEnable, err := shouldEnableScheduler()
 	if err != nil {
 		log.Panic(err)
 	}
 
 	if shouldEnable {
-		crontabWasAdded, err := scheduler.Enable(nil)
+		schedulerEnabled, err := scheduler.Enable()
```
paulfurley: I think dropping the `Was` word lost some nuance. Further on in the code, `schedulerEnabled` could easily be taken as "is the scheduler enabled?". But it doesn't mean that, it means "Was it disabled before, and now we enabled it?"

fk/main.go
```diff
@@ -144,57 +144,63 @@ Options:
...
-		if crontabWasAdded {
-			printInfo(fmt.Sprintf("Added Fluidkeys to crontab.  Edit %s to remove.",
-				Config.GetFilename()))
+		if schedulerEnabled {
+			printInfo(fmt.Sprintf("Added Fluidkeys to %s.  Edit %s to remove.",
+				scheduler.Name(), Config.GetFilename()))
 		}
 	} else {
-		crontabWasRemoved, err := scheduler.Disable(nil)
+		schedulerDisabled, err := scheduler.Disable()
```
paulfurley: Ditto the loss of `Was`

scheduler/cron.go
```diff
@@ -0,0 +1,198 @@
...
+	"strings"
+)
+
+type cron struct{}
+
+// Enable writes Fluidkeys' cron lines into crontab
+func (c *cron) Enable() (crontabWasAdded bool, err error) {
+	crontab := &systemCrontab{}
+	crontabWasAdded, err = c.enable(crontab)
+	if err != nil {
```
paulfurley: When we discussed it, `Enable` did exactly 1 thing:

```
return c.enable(&systemCrontab{})
```

By adding more stuff (wrapping the error check on `enable`), `Enable` is no longer trivial: it has its own behaviour.`

scheduler/cron.go
```diff
@@ -0,0 +1,198 @@
...
+		return crontabWasAdded, ErrModifyingCrontab{origError: err}
```
paulfurley: The intent of the custom `ErrModifyingCrontab` error is to tell the caller there was a problem **modifying** the crontab, so they can go and do it manually.

By wrapping *all errors* from `enable` this specific error can now be raised for numerous reasons, some of which might not be resolved by manually editing your crontab.

Suggest putting `Enable` back to the simple one-liner and returning `ErrModifyingCrontab` from inside `enable`, *only* when it's appropriate.

scheduler/cron.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (c *cron) Name() string {
+	return "crontab"
+}
+
+func (c *cron) enable(crontab runCrontabInterface) (crontabWasAdded bool, err error) {
+	currentCrontab, err := crontab.get()
+	if err != nil {
+		return false, fmt.Errorf("error getting crontab: %v", err)
```
paulfurley: This error, for example, should probably not return `ErrModifyingCrontab`

fk/main.go
```diff
@@ -167,6 +168,25 @@ func ensureSchedulerStateMatchesConfig() {
 				out.Print("\n" + scheduler.CronLines)
 				out.Print(formatFileDivider("", 80))
 				out.Print("\n\n")
+
+			case scheduler.ErrModifyingLaunchd:
```
paulfurley: Is this likely? This feels very fishy!

fk/main.go
```diff
@@ -205,6 +238,10 @@ func ensureSchedulerStateMatchesConfig() {
 	}
 }
 
+func launchdFilename() string {
```
paulfurley: it's a very bad smell that `fk/main.go` knows anything about the internal implementation details of the launchd scheduler!

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+	"os/exec"
+	"path"
+
+	homedir "github.com/mitchellh/go-homedir"
+)
+
+type launchd struct{}
+
+// Enable creates the launchd script and then loads it using launchctl
+func (ld *launchd) Enable() (launchdWasLoaded bool, err error) {
```
paulfurley: As before, if the purpose of `Enable` is just to instantiate `fileFunctionsPassthrough{}` so we can run tests against `enable`, then that's all it should do.

In this case, the functionality seems to be 50/50 split between `Enable` and `enable` !

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+	LaunchAgentsSubdirectory = path.Join("Library", "LaunchAgents")
+	// LaunchAgentFilename is the filename for the launchd agent
+	LaunchAgentFilename       = "com.fluidkeys.launchd.plist"
+	errCouldntFindLaunchdFile = errors.New("couldn't find launchd file")
+)
+
+// ErrModifyingLaunchd is a custom error. It should be used to wrap any errors called during
+// Enable or Disable to ensure the caller knows how to present the user with the steps they
+// can take to manually recify the situation.
+type ErrModifyingLaunchd struct {
```
paulfurley: With crontab we know there's a very specific case we're looking for: where you call the `crontab` command to update the crontab, and due to permissions, it fails.

What's the equivalent here?

Is writing a `launchd` file any more likely to fail than writing, say, our `db.json` file? (We don't make special instructions for the user to manually edit their config file...)

Did you implement this just for symmetry or is there some other way launchd can fail that I haven't spotted yet?

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+        </array>
+        <key>StartInterval</key>
+        <integer>3600</integer>
+    </dict>
+</plist>
+`
+
+var (
+	// LaunchAgentsSubdirectory is the subdirectory for agent files, from the user's home directory
+	LaunchAgentsSubdirectory = path.Join("Library", "LaunchAgents")
```
paulfurley: The comment says it's inside the user's *home directory*, but the code doesn't ...

The path `Library/LaunchAgents` is *not* equal to `/home/user/Library/LaunchAgents`... for a demo, try changing directory into `/` then running `fk` again...

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (ld *launchd) Name() string {
+	return "launchd"
+}
+
+func (ld *launchd) enable(fileFunctions fileFunctionsInterface, launchdAgentFilename string) (
+	launchdFileWasCreated bool, err error) {
+
+	if _, err := fileFunctions.OsStat(launchdAgentFilename); os.IsNotExist(err) {
```
paulfurley: and if `err` is something else, like `permission denied` ?

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	if err := fileFunctions.OsRemove(launchdAgentFilename); err != nil {
+		return false, fmt.Errorf("failed to remove %s: %v", launchdAgentFilename, err)
+	}
+
+	return true, nil
+}
+
+func (ld *launchd) GetAgentFilename() (string, error) {
```
paulfurley: I'm sure your linter said it first, but any reason this is exported? I can only see it used inside this file.

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+	log.Printf("Running `%s %s %s`", launchctl, verb, filename)
+	cmd := exec.Command(launchctl, verb, filename)
+
+	out, err := cmd.CombinedOutput()
+
+	outString := string(out)
+	if err != nil {
+		return outString, err
+	}
+	if outString == filename+": No such file or directory\n" {
```
paulfurley: safer to use `if strings.Contains(outString, "No such file or directory")`

paulfurley: This implies that if you call `launchctl <non existent file>` it returns a *successful* exit code (hence above `err` is nil`) but yet prints out "No such file or directory"... is that true?

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (ld *launchd) runLaunchctl(verb string, filename string) (string, error) {
+	log.Printf("Running `%s %s %s`", launchctl, verb, filename)
+	cmd := exec.Command(launchctl, verb, filename)
+
+	out, err := cmd.CombinedOutput()
+
+	outString := string(out)
+	if err != nil {
```
paulfurley: It's *reeeeally* helpful for our debugging if you dump the full output to log in these cases: 
```
log.Printf("launchd failed with error %v, full output:\n%s", err, outString)
```

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+	return true, nil
+}
+
+func (ld *launchd) GetAgentFilename() (string, error) {
+	homeDirectory, err := homedir.Dir()
+	if err != nil {
+		return "", err
+	}
+
+	return path.Join(homeDirectory, LaunchAgentsSubdirectory, LaunchAgentFilename), nil
```
paulfurley: OHHH here we join with `homeDirectory`. I see.

scheduler/launchd.go
paulfurley: If it turns out the fancy custom error is unnecessary, don't forget to unexport these

fk/main.go
idrysdale: Agreed, I'll reintroduce that.

scheduler/launchd_test.go
```diff
@@ -0,0 +1,109 @@
...
+			OsStatReturnError:          nil,
+			IoutilWriteFileReturnError: nil,
+		}
+		launchdEnabled, err := ld.enable(&mockFileHelper, "fake.plist")
+		assert.NoError(t, err)
+
+		assert.Equal(t, true, launchdEnabled)
+	})
+
+	t.Run("load successfully if file is missing", func(t *testing.T) {
```
paulfurley: does "load successfully if file is missing" mean "create file if it's missing" ?

scheduler/launchd_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+		assert.Equal(t, true, launchdEnabled)
+
+		t.Run("writes out default launchd content with correct mode", func(t *testing.T) {
+			assert.Equal(t, LaunchdFileContents, string(mockFileHelper.IoutilWriteFileGotData))
+			assert.Equal(t, os.FileMode(0644), mockFileHelper.IoutilWriteFileGotMode)
+		})
+	})
+
+	t.Run("error if file is missing and couldn't be created due to permission error", func(t *testing.T) {
```
paulfurley: cool!

scheduler/launchd_test.go
```diff
@@ -0,0 +1,109 @@
...
+			IoutilWriteFileReturnError: nil,
+		}
+		launchdEnabled, err := ld.enable(&mockFileHelper, "fake.plist")
+		assert.NoError(t, err)
+
+		assert.Equal(t, true, launchdEnabled)
+
+		t.Run("writes out default launchd content with correct mode", func(t *testing.T) {
+			assert.Equal(t, LaunchdFileContents, string(mockFileHelper.IoutilWriteFileGotData))
+			assert.Equal(t, os.FileMode(0644), mockFileHelper.IoutilWriteFileGotMode)
```
paulfurley: You're explicitly testing that it should be written out with `0644` access permissions ... why those permissions?

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+	return "launchd"
+}
+
+func (ld *launchd) enable(fileFunctions fileFunctionsInterface, launchdAgentFilename string) (
+	launchdFileWasCreated bool, err error) {
+
+	if _, err := fileFunctions.OsStat(launchdAgentFilename); os.IsNotExist(err) {
+		// file does not exist, write out default launchd agent file
+		if err = fileFunctions.IoutilWriteFile(
+			launchdAgentFilename, []byte(LaunchdFileContents), 0644); err != nil {
```
paulfurley: `0644` ... why?

scheduler/launchd_test.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type mockFileFunctions struct {
+	// provides fake versions of os.Stat etc.
+	// implements fileFunctionsInterface
+
+	OsRemoveReturnError        error
+	OsStatReturnError          error
+	IoutilWriteFileReturnError error
+	FileContents               string
```
paulfurley: Is `FileContents` used?

scheduler/scheduler.go
```diff
@@ -17,10 +17,17 @@
 
 package scheduler
 
+import "os/exec"
+
 var scheduler schedulerInterface
 
 func init() {
-	scheduler = &cron{}
+	_, err := exec.LookPath("launchd")
```
paulfurley: `LookPath` ! cool!

paulfurley: Hmm, given that the executable you actually run is called `launchctl`, wouldn't it make sense to look for that? (It's even defined: `const launchctl string = "launchctl"`

scheduler/launchd.go
```diff
@@ -0,0 +1,149 @@
...
+
+const launchctl string = "launchctl"
+
+// LaunchdFileContents is the agent file for running fk sync every 60 minutes
+const LaunchdFileContents = `<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+    <dict>
+        <key>Label</key>
+        <string>com.fluidkeys.sync</string>
```
paulfurley: should this match the filename (`com.fluidkeys.launchd.plist`) ?

scheduler/launchd_test.go
idrysdale: I looked at what the file permissions of all the other `.plist` files I have in that directory are for other applications - they are `-rw-r--r--`

scheduler/scheduler.go
idrysdale: Agreed!

scheduler/cron.go
idrysdale: Yep - you're right.

scheduler/cron.go
idrysdale: Yep, stripped back.

scheduler/cron.go
idrysdale: 👍

scheduler/launchd.go
idrysdale: Good question. I was taking my lead from `cron`, so for symmetry. In hindsight, the only way it's going to fail is if it can't write that file to disk.

paulfurley: *Can* `launchctl` fail? Does it have some permissions itself?

scheduler/launchd_test.go
paulfurley: Good man!

Even so, I suggest to use `600`, I don't see why we want any other users being able to read the file.

scheduler/launchd.go
idrysdale: As far as I can tell...
```
launchctl load nonexistant.plist
/Users/idrysdale/go/src/github.com/fluidkeys/fluidkeys/nonexistant.plist: No such file or directory
➜  fluidkeys git:(add-launchd) ✗ echo $?
0
```

scheduler/launchd.go
idrysdale: 👍

scheduler/launchd.go
paulfurley: Absolutely right! It's worth a comment there to document `luanchctl`'s (weird) error behaviour

scheduler/launchd_test.go
idrysdale: Roger.

scheduler/launchd_test.go
idrysdale: Nope!

scheduler/launchd.go
idrysdale: `os.Stat` can only return errors of `*PathError` - this line is simply checking if the file exists to see whether we need to write it.

If there's a permission error, I presume it'd be caught by `launchctl`?

scheduler/launchd.go
idrysdale: Yes, that's probably a good idea!

scheduler/launchd.go
idrysdale: Modified this, factoring out a function `fileExists`.

paulfurley:
Cool!

Probably final thing then, how're we going to migrate from cron to launchd (e.g. remove existing Fluidkeys cron lines?)

PR #483, opened 2019-05-01
team fetch: getUnlockedKey: if already unlocked, don't prompt

paulfurley:
the first time we successfully unlock a key, cache the result so we
don't need to unlock it again in subsequent calls.


idrysdale:
Neato!

PR #484, opened 2019-05-01
team fetch: don't unlock key unless required

paulfurley:
only unlock the key if we actually need it to certify a new key

(I'm getting blocked by all these dependency pull requests)

### DIFF
--- a/fk/teamfetch.go
+++ b/fk/teamfetch.go
@@ -61,28 +61,16 @@ func teamFetch(unattended bool) exitCode {
 }
 
 func doUpdateTeam(myTeam *team.Team, me *team.Person, unattended bool) (err error) {
-	alwaysDownload := !unattended
-
 	printHeader(myTeam.Name)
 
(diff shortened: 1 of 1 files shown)


fk/teamfetch.go
```diff
@@ -61,28 +61,16 @@ func teamFetch(unattended bool) exitCode {
...
-			"Failed to unlock key to check for team updates", []string{
-				"Checking for updates to the team requires an unlocked key",
-				"as the team roster is encrypted.",
-			}, err))
-		return err
-	}
-
 	var updatedTeam *team.Team
-	if updatedTeam, err = fetchAndUpdateRoster(*myTeam, unlockedKey, alwaysDownload); err != nil {
+	if updatedTeam, err = fetchAndUpdateRoster(*myTeam, *me, unattended); err != nil {
```
idrysdale: In a previous commit, I remember `me` was a fingerprint. I wonder whether we should just be passing a `fingerprint` to the functions in this file, rather than a `Person`. Thoughts?

PR #488, opened 2019-05-02
team fetch: log success and failure in unattended mode

paulfurley:
this is to give us visibility of whether the fk client is running reliably
from the scheduler for teams.

### DIFF
--- a/fk/teamfetch.go
+++ b/fk/teamfetch.go
@@ -47,8 +47,26 @@ func teamFetch(unattended bool) exitCode {
 	}
 
 	for i := range memberships {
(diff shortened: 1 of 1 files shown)


fk/teamfetch.go
```diff
@@ -47,8 +47,26 @@ func teamFetch(unattended bool) exitCode {
...
+				api.Log(apiclient.Event{
+					Name:        eventErrorUpdatingTeamUnattended,
+					Fingerprint: &me.Fingerprint,
+					TeamUUID:    &t.UUID,
+					Error:       err,
+				})
+			}
+		} else if unattended {
+			api.Log(apiclient.Event{
+				Name:        eventErrorUpdatingTeamUnattended,
```
idrysdale: This should be successful log!

PR #492, opened 2019-05-03
after `fk team apply`, keep checking

paulfurley:
every 30 seconds to see if we've been added to the team yet.

after 4 hours, give up.

either way, run `fk team fetch` afterwards

### DIFF
--- a/fk/teamapply.go
+++ b/fk/teamapply.go
@@ -19,17 +19,21 @@ package fk
 
 import (
 	"fmt"
+	"log"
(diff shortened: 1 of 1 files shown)


idrysdale:
This is strange. I've tested it twice, and both times after having had their request to join approved, the requester gets this:

Screenshot 2019-05-03 at 14 53 27

idrysdale:
Doh! I was using an out-of-date `api`! 🤦‍♂

PR #493, opened 2019-05-03
refer to correct scheduler during `key maintain`

idrysdale:


paulfurley:
Awesome! There are several more of these...

(as mentioned in [this trello card](https://trello.com/c/9hw5cG2A)

PR #494, opened 2019-05-03
enable re-joining a team you've deleted

paulfurley:
if you've trashed your `teams/team` subdirectory, previously it was difficult to rejoin a team.

Assumption: you're listed in the team roster on the API.

Scenarios:

1. no requests to join team in local DB or in API
2. got a request to join team in local DB but *not* in the API
3. no request to join team in local DB but *got one* in the API
4. got request to join team in local DB and in the API

If you've got a local request already, it presents an infobox with your verification details
you need to give the admin.

For this scenario the solution is running `fk team fetch`. I've moved the call to action (run `fk
team fetch`) to the bottom so it's clearer.

tricker because the API rejects your request to join the team, saying
'Failed to apply: Already got request'

Now, we store a request in the db, ask the API if we're already in the team, and if so,
pass straight off to `fk team fetch`, which will pick up the request from the db and join the
team.

We skips requesting to join in the API, so don't hit that issue.

As with 3, we store a local request, ask the API, then pass straight through to `fk team fetch`

### DIFF
--- a/fk/teamapply.go
+++ b/fk/teamapply.go
@@ -67,7 +67,24 @@ func teamApply(teamUUID uuid.UUID) exitCode {
 
 	printHeader("Apply to join team")
 
-	if err := requestToJoinTeam(teamUUID, teamName, pgpKey.Fingerprint(), email); err != nil {
+	// always record a request so `team fetch` can refer to it later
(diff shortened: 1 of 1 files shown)


idrysdale:
Excellent!

PR #495, opened 2019-05-03
Instruct to allow keychain access

idrysdale:
We do this on both `key create` and `key from-gpg`, but only if the keyring
backend is KeyChain.

This is how it looks on macOS.

For *key create*:
Screenshot 2019-05-03 at 15 18 57

For *key maintain*:
Screenshot 2019-05-03 at 15 20 29

### DIFF
--- a/fk/keymaintain.go
+++ b/fk/keymaintain.go
@@ -345,6 +345,21 @@ func promptAndTurnOnMaintainAutomatically(prompter promptYesNoInterface, keyTask
 	out.Print("This requires storing the password in the system keyring.\n\n")
 
 	if prompter.promptYesNo(promptMaintainAutomatically, "", keyTask.key) == true {
+
+		out.Print("Fluidkeys must be able to read back the password from " +
+			Keyring.Name() + "\n")
+		if Keyring.PermissionsInstructions() != "" {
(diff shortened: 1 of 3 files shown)


fk/keymaintain.go
```diff
@@ -345,6 +345,19 @@ func promptAndTurnOnMaintainAutomatically(prompter promptYesNoInterface, keyTask
 	out.Print("This requires storing the password in the system keyring.\n\n")
 
 	if prompter.promptYesNo(promptMaintainAutomatically, "", keyTask.key) == true {
+		if Keyring.ShouldInstructAboutPermissions() {
```
paulfurley: Why only on macOS? I'd be perfectly happy to see it read back from my linux keyring, even if I didn't have to click "allow"

paulfurley: I'm also not sure what other flavours of keyring are out there. Maybe on Arch it would prompt?

idrysdale: I've modified this now and force pushed!

paulfurley:
looks great here!
image

paulfurley:
BTW we should totally try and use `RunWithCheckboxes` in this file sometime!

PR #504, opened 2019-05-09
`team apply`: don't get stuck unable to join a team

paulfurley:
previously it was possible that the local DB has a request to join a team, but the API doesn't

that makes it impossible to re-request to join the team, *and* for an admin to authorize the person

these changes:
1. make it less likely the mismatch will occur in the first place
2. make it recoverable if that does occur

### DIFF
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -139,20 +139,20 @@ func TestRecordRequestsToJoinTeamG(t *testing.T) {
 		})
 	})
 
-	t.Run("record deduplicates selecting oldest requests for team & fingerprint", func(t *testing.T) {
+	t.Run("record deduplicates selecting newest (latest) requests for team & fingerprint", func(t *testing.T) {
 		database := New(testhelpers.Maketemp(t))
 
-		dupeReq := request1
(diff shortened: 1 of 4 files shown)


database/database_test.go
```diff
@@ -201,41 +201,41 @@ func TestGetRequestsToJoinTeams(t *testing.T) {
...
 		gotRequests, err := database.GetRequestsToJoinTeams()
 		assert.NoError(t, err)
 
-		expectedRequests := []team.RequestToJoinTeam{oldestReq}
+		expectedRequests := []team.RequestToJoinTeam{newerReq1}
 		assert.Equal(t, len(expectedRequests), len(gotRequests))
 		assert.Equal(t, expectedRequests, gotRequests)
 	})
 
 	t.Run("returns oldest when oldest was inserted first", func(t *testing.T) {
```
idrysdale: I think this test should read "returns *newest* when *newest* was inserted first"

idrysdale:
Other than that test name, this is good to go! 👍

database/database_test.go
paulfurley: I've nested the two cases to make it clearer

PR #507, opened 2019-05-10
remove hardcoded kiffix teamname

idrysdale:
🤦


fk/teamapply.go
```diff
@@ -96,7 +96,7 @@ func teamApply(teamUUID uuid.UUID) exitCode {
 		"This information allows them to verify your request.",
 	}))
 
-	out.Print(formatFileDivider("Please authorize me to join Kiffix", 80) + "\n")
```
paulfurley: !!! OMG how long as his been in there for?!

